PR #4, opened 2023-11-27
Add the new mal compiler

nkakouros:
I added an new subpackage `maltoolbox.lexer_parser`. I added all related files in its folder, we can move them around as needed. The code is minimal, it contains:

- the grammar
- antlr-generated files based on the grammar
- a tree parser module
- a runner script with just one argument required and supported

To run the compiler, you run `python -m maltoolbox.lexer_parser path/to/corelang_or_other/main.mal`.


andrewbwm:
This was manually squashed and merge into main.

Issue #6, opened 2024-01-24
Implement compiler checks


nkakouros:
The original mal compiler implemented checks for various aspects of syntax validity, constraints, etc. These can be found [here](https://github.com/mal-lang/malcompiler/blob/master/malcompiler-lib/src/main/java/org/mal_lang/compiler/lib/Analyzer.java).

### TODO
- [ ] Study the mal language. Here are some resources:
  - [ ] https://github.com/mal-lang/mal-documentation/wiki
  - [ ] https://github.com/mal-lang/exampleLang/
- [ ] Understand the new mal compiler:
  - [ ] Study the mal syntax grammar, see [this PR](https://github.com/mal-lang/mal-toolbox/pull/4/files) and the `.g4` grammar file, until it gets merged.
  - [ ] Understand the visitor pattern of the compiler
- [ ] Implement the validator checks
  - [ ] Understand each check in the original compiler's analyzer
  - [ ] Implement each check in the original compiler's analyzer

andrewbwm:
It might be useful for the person doing this to also have a look at the Language Graph(https://github.com/mal-lang/mal-toolbox/blob/main/maltoolbox/language/languagegraph.py).

PR #8, opened 2024-02-12
Analyzer/Checker draft

WLUB:
# Analyzer/Checker draft

This is a draft for the analyzer/checker described in issue #6   
Only some of the checks are implemented into this version. 

## Implementation
The analyzer is a parent to the compiler `mal_visitor.py`.   
For every visit state we define inside `mal_visitor.py`, we also define inside `mal_analyzer.py` (if we want to check it). We can then access the `mal_analyzer.py` visit-method via super(...).
To make the code more readable there is a `decoration`  class inside `malAnalyzer` that will implement the super(...) code into the original visit method and also trigger an error if the method isn't defined inside `mal_analyzer.py `.

`ParseTreeVisitor` method `visit` is overridden in order to preform a `post-analysis` as the last step. `post-analysis` is for checks where we need to iterate all the assets or similar collections of data. By using the visit method to trigger `post-analysis` we don't need to change how the compiler is used. 

With this implementation it is very easy to add new rules to analyze/check and the core compiler only needs to add one line of code for every new `visitor` that should be checked.

## Naming

- Use the name `analyzer` or `checker`?
- Call the last step `post-analysis`?

## Potential issues / changes

- The `visit` method in `mal_analyser` is doing more then the method is meant to do, which can lead to confusion.
- The `Analyze` decoration can be removed in order to use super(...).visit(...) instead. Downside is no warnings for non-existing super methods.
- Don't have `malAnalyzer ` as parent to `malVisitor`, we could make `malAnalyzer` as a "normal" class and then preform calls in every visit function instead. 
- Some of the methods should be refactored.

### DIFF
--- a/maltoolbox/language/lexer_parser/__main__.py
+++ b/maltoolbox/language/lexer_parser/__main__.py
@@ -3,7 +3,7 @@
 from antlr4 import FileStream, CommonTokenStream
 from .mal_lexer import malLexer
 from .mal_parser import malParser
-# from .mal_visitor import malVisitor
+from .mal_visitor import malVisitor
 from .mal_analyzer import malAnalyzer
 
 import sys
@@ -30,7 +30,8 @@ class MalCompiler:
         parser = malParser(stream)
         tree = parser.mal()
 
-        return malAnalyzer(compiler=self).visit(tree)
+        analyzer = malAnalyzer()
+        return malVisitor(compiler=self, analyzer=analyzer).visit(tree)
 
 
 if __name__ == "__main__":
--- a/maltoolbox/language/lexer_parser/__main__.py
+++ b/maltoolbox/language/lexer_parser/__main__.py
@@ -3,7 +3,8 @@
 from antlr4 import FileStream, CommonTokenStream
 from .mal_lexer import malLexer
 from .mal_parser import malParser
-from .mal_visitor import malVisitor
+# from .mal_visitor import malVisitor
+from .mal_analyzer import malAnalyzer
 
 import sys
 import os
@@ -29,7 +30,7 @@ class MalCompiler:
         parser = malParser(stream)
         tree = parser.mal()
 
-        return malVisitor(compiler=self).visit(tree)
+        return malAnalyzer(compiler=self).visit(tree)
 
 
 if __name__ == "__main__":
--- a/maltoolbox/language/lexer_parser/mal_analyzer.py
+++ b/maltoolbox/language/lexer_parser/mal_analyzer.py
@@ -1,0 +1,235 @@
+from antlr4 import ParseTreeVisitor
+from .mal_parser import malParser
+
+from dataclasses import dataclass
+import logging
+import re
+    
+class malAnalyzer(ParseTreeVisitor):
+    '''
+    A class to preform syntax-checks for MAL.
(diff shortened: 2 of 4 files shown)


nkakouros (commented):
It's great that we now have an implementation; it will help with deciding which approach works best. You also seem to have a good grip on the language and what needs to be checked. Here are some comments:

The `visit` method is the operational, the visiting method. It is supposed to be used when visiting a context (instead of the implementing method, e.g.`visitMal` or `visitStep`). Checking the depth in `visit` to kickstart the analyzer works but it goes against the single responsibility principle. I think a better place would have been the visitMal method that already operates on depth 1. 

The use of the `malAnalyzer.analyze` decorator is clever. But does it provide significant savings? Instead of having a `super()...` line, we now have a decorator line for each overridden method and a decorator class definition that makes calling the parent's implementation more obscure.

I also think that for the mal visitor to inherit from the analyzer seems a bit weird. On one hand conceptually it makes sense to me to be the other way around. On the other, some operations need to be replicated this way. For example, in `visitDefine` in the analyzer class you have:

```
define_id: str = ctx.ID().getText()
```

The same code exists in the `visitDefine` method of the mal visitor class. This should not affect performance but it make the code repeatable. 

Instead, you could have the analyzer inherit from the mal visitor and when overriding a method to add checks to it you could do:

```
result = super().visitDefine(...)

_, define = result  # result is a tuple with `"defines"` as the first item and a single-item dict as the second

if next(iter(define.values())) is None:  # single item in the `define` dict
  # fail here due to empty define value
```

I think this would look cleaner. We had discussed it as a potential approach in an email exchange, but seeing a first implementation now I think it makes more sense to do it this way.

Also, as you have seen, some checks about an entity, e.g. defines, may not be possible to place under its visit method. For instance, to check if there is a define that gives an `id` we need to check the whole mal spec. As mentioned above, such checks could run in the `visitMal` method of the analyzer class that extends the mal visitor, as this visitor method is at depth 1. 

An alternative could be to just use the mal visitor as is, get a tree out of it and then just apply the checks on the tree (the dict that contains all the parsed nodes). But we will not be able to stop compiling as soon as an error can be detected (for instance, the moment when we see that a define does not have a value).

A slight problem with having the analyzer inherit from the visitor is that if you want to compile sth programmatically, you need to use the analyzer class which is counter-intuitive. To counter that, we could use dependency injection instead of inheritance, where we pass an Analyzer object to the visitor and the visitor uses it where it makes sense, as described above. I think this would be the cleanest approach. E.g. in `malVisitor`'s `visitDefine` we would call `self.analyzer.check_empty_define(the_value_to_be_returned_here)`. Or we could keep all checks in the `visitMal` method:

```python
        for declaration in (d.getChild(0) for d in ctx.declaration()):
            if result := self.visit(declaration) or True:
                key, value = result

                if key == "categories":
                    ...

                if key == "defines":
                    self.analyzer.check_emtpy_define(value)
                    # or self.analyzer.check_define(value)  # where multiple checks on the single define can happen
                    # and a self.analyzer.check_defines() method would run later in visitMal once
                    # the whole tree is available to check sth on the whole set of defines.

                    langspec[key].update(value)
```

nkakouros:
Superseded by #111.

PR #10, opened 2024-03-18
Simplifications and pythonic improvements

nkakouros:
No new functionality here, just simplifications and making some of the code more pythonic. Given that the ambition is for these libraries to be used outside of our small group, it should be worth it to consider the Python-ism of the code. I have split the PR into quite contained commits to better review it (commit-by-commit).


maltoolbox/language/languagegraph.py
```diff
@@ -19,9 +19,9 @@
...
     name: str = None
-    associations: list[LanguageGraphAssociation] = None
-    attack_steps: list[LanguageGraphAttackStep] = None
-    description: dict = None
+    associations: list[LanguageGraphAssociation] = field(default=None, repr=False)
+    attack_steps: list[LanguageGraphAttackStep] = field(default=None, repr=False)
+    description: dict = field(default=None, repr=False)
     # MAL languages currently do not support multiple inheritance, but this is
     # futureproofing at its most hopeful.
     super_assets: list = None
```
kasanari, line 27: The fields that are initialized to None should either be marked with Optional or initialized to empty dicts/lists with field()

andrewbwm:
This was included as part of the [0.0.29 release](https://github.com/mal-lang/mal-toolbox/commit/fc7158d95dc18b64c03c6969177b2db59b891460).

PR #13, opened 2024-03-27
Make LanguageGraph source of truth for DSL + various

nkakouros:
This PR attempts to make the LanguageGraph a required argument for any other class that needs access to the DSL. There are several TODOs in the code that we should discuss and resolve or convert to issues before merging.

Due to the hard parts of constructing the attack graph, this PR does not attempt to use the LanguageGraph for attack graph construction. Still, it passes the LanguageGraph to the `AttackGraph` constructor and reads the `lang_spec` (the DSL) through the `LanguageGraph`. Ideally, this should not be needed, and `LanguageGraph` properties and methods should be used even for the attack graph construction. Implementing that will surely reveal some common functionality between the LanguageGraph and some parts of the `AttackGraph._generate()` method (and potentially others). Due to the complexity of the code, @andrewbwm should lead this effort, I believe.

Also, this PR suggests that we deprecate the `language.specification` module and move that code elsewhere for better discoverability and encapsulation. Some parts have already been moved, but there are still a few functions in that module.

The PR also distinguishes between (unqualified) `name` and `qualified_name` for the steps in the LanguageGraph. The same can be done for the nodes in the attack graph. This will make clear what is being accessed each time and will also contain the concatenation of `:` and other parts in a single point in code. This will also help with the `unholy` thing in the simulator, making it much simpler to achieve what that option is supposed to do.

Other commits in this PR should be self-explanatory.

If and when this gets merged, the mal simulator should also be updated.


maltoolbox/language/classes_factory.py
```diff
@@ -12,8 +12,9 @@ class LanguageClassesFactory:
     def __init__(self, lang_graph):
         self.lang_graph = lang_graph
         self.json_schema = {}
+        self.__create_classes()
 
-    def generate_assets(self):
+    def __generate_assets(self):
```
nkakouros, line 17: A minor thing, prepending with 2 underscores will cause name mangling. This method will be accessible via:

```
class_factory_instance._LanguageClassesFactory__private_method()
```

Is there indeed a reason to prefer this over the single underscore convention for private methods?

andrewbwm: No, I just wasn't aware which of the two was preferred, I will change it in a future commit.

PR #37, opened 2024-05-23
Yaml from languagegraph and attackgraph

mrkickling:


nkakouros:
Sth that I noticed when working with an instance model of about 2M lines is that yaml is 10x or more slower to parse the yml file. A JSON version of the same file loads in 2 seconds instead.

PR #41, opened 2024-05-31
Optimizations

mrkickling:


maltoolbox/attackgraph/attackgraph.py
```diff
@@ -41,11 +41,11 @@ def _process_step_expression(lang_graph: LanguageGraph, model: Model,
...
     """
-    logger.debug('Processing Step Expression:\n' \
-        + json.dumps(step_expression, indent = 2))
 
-
-    lang = lang_graph._lang_spec
+    if logger.level == logging.DEBUG:
+        # This expressions takes a lot of time
+        logger.debug('Processing Step Expression:\n' \
+            + json.dumps(step_expression, indent = 2))
```
nkakouros, line 48: The [suggested way](https://docs.python.org/3/howto/logging.html#optimization) to do this is:

```
if logger.isEnabledFor(logging.DEBUG):
    ...
```

Also, the suggested way to format strings in log statements is with `%s` and passing the variables as arguments, like `debug("string %s %s", arg1, arg2)`, e.g.:

```
     logger.debug('Processing Step Expression:\n%s',
            json.dumps(step_expression, indent = 2))
```

So that the string gets formatted only if the debug level is enabled. The dumps call, however, will still be evaluated regardless of the debug level and this is why you need to conditional check anyway.

But adding the conditional check all the time (given that there are a lot of debug statements in the code) seems cumbersome to me. I would create a custom logging formatter, use the suggested way to format log strings and dump to json in the formatter's `format()` method only if an argument is a dict. Or, I would send a keyword arg along side the positional args and check that for dumping to json, e.g.:

```
debug("string %s %s", str_var1, str_var2, json=True)
```

maltoolbox/attackgraph/attackgraph.py
```diff
@@ -165,7 +165,9 @@ class AttackGraph():
     """Graph representation of attack steps"""
     def __init__(self, lang_graph = None, model: Optional[Model] = None):
         self.nodes = []
+        self.id_to_node = {}  # to get nodes by id faster
```
nkakouros, line 168: I think there is a bigger issue in the current implementation where lists are preferred over dicts. This forces the code to search in lists if e.g. a specific assets needs to be found. It would make the code both simpler and faster if dicts where used that were based by the asset id or name, association name, etc. Maybe it is worth rewriting the toolbox around dicts?

maltoolbox/model.py
```diff
@@ -53,28 +58,27 @@ def add_asset(
...
-                if ex_asset.name == asset.name:
-                    if allow_duplicate_names:
-                        asset.name = asset.name + ':' + str(asset.id)
-                        break
-                    else:
-                        raise ValueError(f'Asset name {asset.name} is a '
-                        'duplicate and we do not allow duplicates.')
+            if asset.name in self.asset_names:
+                if allow_duplicate_names:
+                    asset.name = asset.name + ':' + str(asset.id)
```
nkakouros, line 77: Irrelevant but the docstring says the name will be prefixed with the id, not suffixed.

nkakouros:
I think before going into the details of each line of code a bigger gain in performance would be made if the code was de-duplicated and streamlined further. For example, `maltoolbox.language.specification` should be removed. The first function in there:

https://github.com/mal-lang/mal-toolbox/blob/9acd22f77028df556f00abbadf9b39bc516e5848/maltoolbox/language/specification.py#L12-L17

is used only by securicad.py which seems unmaintained anyway. 

Then, there is:

https://github.com/mal-lang/mal-toolbox/blob/9acd22f77028df556f00abbadf9b39bc516e5848/maltoolbox/language/specification.py#L55-L68

which replicates (in a bad way) the functionality of:

https://github.com/mal-lang/mal-toolbox/blob/9acd22f77028df556f00abbadf9b39bc516e5848/maltoolbox/language/languagegraph.py#L61-L72

Removing these two functions will remove specifications.py too.

Then, there is duplication of functionality at least in one place. The following method:

https://github.com/mal-lang/mal-toolbox/blob/9acd22f77028df556f00abbadf9b39bc516e5848/maltoolbox/language/languagegraph.py#L443-L468

has a counterpart in attackgraph.py:

https://github.com/mal-lang/mal-toolbox/blob/9acd22f77028df556f00abbadf9b39bc516e5848/maltoolbox/attackgraph/attackgraph.py#L26-L42

Even if slight changes exist between the two, they should be merged as they deal with the same task of processing step expressions.

Then, the attack graph generation code does not really use the language graph and its methods to do its job. If it did, some code would be removed and simplified.

Although all of the above may not yield performance gains, I believe they are prerequisites to optimizing the code effectively.

Given that the above will be a medium refactoring of the code, I believe it is worth rewriting the internals around dicts rather than lists to get some quick speed gains before diving into actual profiling.

maltoolbox/attackgraph/attackgraph.py
mrkickling: Thanks, I will change the if statement.
Yes, I think all of our logging statements to use the %s-method instead, but I think it should be a future PR for that. Same with your suggestion on a custom formatter. Do you agree we can do it at a later time?

maltoolbox/attackgraph/attackgraph.py
mrkickling: Yes, maybe we can replace the list all together even!
Also think this should be a new issue/PR because it might be many changes to the code unrelated to optimizations.

maltoolbox/model.py
mrkickling: Thanks!

maltoolbox/attackgraph/attackgraph.py
nkakouros: A potential issue with deferring that I see is that this task may change the API of the toolbox and introduce breaking changes. But you know what priorities are there.

nkakouros:
Specifically for `process_step_expression`, do you think it will be hard to use only the LanguageGraph version (and add any missing bits there from the attackgraph.py version)? That would be a great first step towards using the lang_graph for attackgraph generation.

maltoolbox/attackgraph/attackgraph.py
mrkickling: Created PR for using format strings in logging, working on it #48

maltoolbox/attackgraph/attackgraph.py
mrkickling: Could be an issue if someone accesses .assets or .nodes, that is true..

PR #43, opened 2024-06-04
Remove unused function from neo4j.py

nkakouros:
I couldn't find `get_model` being used anywhere.


nkakouros:
> we do need to maintain more things at the same time

And this results in unmaintained code like the one this PR removes. 

If I understand what `get_model` does, it creates an instance model from a neo4j instance. I think this is clearly out of scope for the mal-toolbox, don't you think?

andrewbwm:
> > we do need to maintain more things at the same time
> 
> And this results in unmaintained code like the one this PR removes.
> 
> If I understand what `get_model` does, it creates an instance model from a neo4j instance. I think this is clearly out of scope for the mal-toolbox, don't you think?

This was not something that was left over from some early iteration, it was something that Giuseppe specifically requested at the time. It was out of scope when it was introduced, the question if it's needed.

nkakouros:
It's your call. My input is that the potential of seeing specication.py removed is great news. If you don't feel comfortable removing the securicad module, perhaps the relevant code from specification.py that is used only in securicad.py could go in there and specification.py is removed?

PR #45, opened 2024-06-04
Fixes subType expression parsing

nkakouros:
The `specification.extends_asset` method is expecting asset names as input while the attack graph uses LanguageGraphAsset objects. 

If this gets merged, `specification.extends_asset` will be used only in securicad.py which is outdated anyway.


andrewbwm:
Implemented in #69 .

PR #46, opened 2024-06-04
Deletes securicad.py and specification.py

nkakouros:
This PR builds on #45 and #43 (only the last commit in this PR is new). 

If these 2 PRs get merged, then if securicad.py is gone, specification.py is not used anymore.

securicad.py is unmaintained. It also does more than what is needed to parse securicad files. If parsing such files is still needed, we can write a **much** simpler transpiler that converts the securicad model format to our own model format.


andrewbwm:
Implemented in #69 .

PR #66, opened 2024-07-11
Don't assume about the virtual env dir name

nkakouros:


nkakouros:
Actually, mypy ignores files/dirs starting with `.`. So why not install your venv under .venv?

mrkickling (approved):
Makes sense, feel free to merge

nkakouros:
Actually, I will make some further edits.

andrewbwm:
> Actually, I will make some further edits.

@nkakouros what is going on with this pull request? Is it still relevant? Is still being worked on? Did the intervening changes make it obsolete or require an update?

nkakouros:
I don't remember. There was a better way to handle some things, but I am not sure anymore. Given that the checks pass and in particular that the mypy check finds and checks all files, I would merge this to avoid virtualenv folders from being picked up by mypy.

andrewbwm:
Was squashed into 2df9b74a77a60206ae631ca368080bfb96f5f86b.

PR #69, opened 2024-07-18
Remove Specification

andrewbwm:
This pull request actually does multiple things:

1. First, as per its name it removes the specification submodule of the language module. The `get_association_by_fields_and_assets` method was moved into the language graph and the `extends_asset`  method was already implemented by `is_subasset_of`.
2. It then updates the securiCAD translator to work with the current version of the model format and structure.
3. It also **partially** updates the Neo4j ingestor module to be able to extract the model from a Neo4j instance. This does not fully work right now, but since nobody has been working with it right now it was deemed a low priority.

1. and 2. should make #46 obsolete.
1.  should make #45 obsolete.


maltoolbox/ingestors/neo4j.py
```diff
@@ -184,41 +184,39 @@ def get_model(
             target_id = right_id
             target_prop = left_field
 
-        if attacker_id:
+        if not attacker_id == None:
```
mrkickling: is not None

maltoolbox/attackgraph/attackgraph.py
```diff
@@ -150,11 +150,28 @@ def _process_step_expression(
                     step_expression['stepExpression'])
                 new_target_assets.extend(assets)
 
-            selected_new_target_assets = [asset for asset in \
-                new_target_assets if specification.extends_asset(
-                    lang_graph._lang_spec,
-                    asset.type,
-                    step_expression['subType'])]
+            selected_new_target_assets = []
+            for asset in new_target_assets:
```
mrkickling, line 154: Just a thought, does a language have assets or should we call them asset_types instead?
Assets are in my understanding part of models, and they are created from the asset types defined in the language.  This is nothing related to this PR in particular, just about the terminology used in the LanguageGraph.

maltoolbox/language/classes_factory.py
```diff
@@ -184,3 +184,59 @@ def _create_classes(self) -> None:
...
+
+        Arguments:
+        assoc_name          - the association name
+        left_asset          - the name of the left asset type
+        right_asset         - the name of the right asset type
+
+        Return: The matching association name if a match is found.
+        None if there is no match.
+        """
+        if not assoc_name in self.json_schema['definitions']\
```
mrkickling, line 207: I would split out self.json_schema['definitions']['LanguageAssociation']['definitions'] into a variable called 'association_definitions' or something, since it is used more than once and increase readability

maltoolbox/translators/securicad.py
```diff
@@ -100,7 +104,7 @@ def load_model_from_scad_archive(
             target_id = right_id
             target_prop = child.attrib['sourceProperty']
 
-        if attacker_id:
+        if  attacker_id is not None:
```
mrkickling, line 107: extra space between attacker_id i think?

maltoolbox/attackgraph/attackgraph.py
andrewbwm: Yes, I had the exact same feeling when working on the language graph changes that the fact that both are called assets is very confusing. I think this will get resolved if we end up using the language graph to create the attack graph. So, it's a good observation, but we will see how it gets resolved if/when we get to that pull request.

mrkickling (approved):
lgtm

nkakouros (approved):

PR #77, opened 2024-08-09
Make sure correct errors thrown when adding attacker with 'bad' entry…

mrkickling:
… points or reached attack steps


andrewbwm:
What was adding entry points as strings? I actually don't like having the entry points either be integers or strings of integers, it kind of muddies the waters for no good reason.

mrkickling:
You are right, we should change it in an earlier stage.
Reverted and created a new solution.

andrewbwm (approved):

PR #78, opened 2024-08-14
Simplify convert old model

mrkickling:
Add a small CLI (maltoolbox.translators.updater) that can be used to convert old models to new ones with just command line.


mrkickling:
No longer relevant after #104 is merged. Closing.

Issue #51, opened 2024-06-11
Attack Graph Does Not Handle 'Collect' Step Expression Correctly


andrewbwm:
Both the attack graph generation incorrectly handle the 'collect' step expression correctly and apply the right hand to all of the assets on the left hand collectively instead of individually resolving it against each one.

This is an issue especially when dealing with intersection and difference operators where the results may be incorrect.

Example: `(a.field1 /\ a.field2)` should be different to `a.(field1 /\ field2)`.

PR #87, opened 2024-10-13
Lang graph integration

nkakouros:
Creating PR to be able to discuss the code. I am not the author of the code changes.

Added by Andrew:
Merge checklist:

- [x] See if #88 was resolved
- [ ] See which of the bugs on [Fixes branch](/github/mal-lang/mal-toolbox/tree/fixes_for_maltoolbox) are still an issue
- [x] See if #51 was resolved


maltoolbox/language/languagegraph.py
```diff
@@ -985,31 +1565,10 @@ def _get_attacks_for_asset_type(self, asset_type: str) -> dict:
...
             'language specification.', asset["name"]
         )
-        if asset['superAsset']:
-            logger.debug('Asset extends another one, fetch the superclass '\
-                'attack steps for it.')
-            attack_steps = self._get_attacks_for_asset_type(asset['superAsset'])
 
         for step in asset['attackSteps']:
             if step['name'] not in attack_steps:
                 attack_steps[step['name']] = copy.deepcopy(step)
```
nkakouros, line 1571: Why not do this:

```suggestion
        attack_steps = copy.deepcopy(asset['attackSteps'])
```

Is there ever a case where an lang-graph asset will contain the same step twice? Shouldn't this be considered a bug at the .mal level?

Also, I don't think deepcopy is needed here. The returned `LanguageGraphAttackStep`'s are only used for lookup of step info when constructing the attack graph.

maltoolbox/language/languagegraph.py
```diff
@@ -116,9 +138,48 @@ def get_all_superassets(self) -> list[LanguageGraphAsset]:
...
+        """
+        Return a list of all of the associations that belong to this asset
+        directly or indirectly via inheritance.
+
+        Return:
+        A list of all of the associations that apply to this asset, either
+        directly or via inheritance.
+        """
+
+        associations = dict(self.associations)
```
nkakouros, line 152: `self.associations` is a list. This will result in an error. For quicker lookup, wouldn't it make sense to have associations be a dict of some sort anyway?

Also, `get_associations` could be converted to an `@property` for a simpler interface to the language graph.

maltoolbox/attackgraph/node.py
```diff
@@ -25,7 +25,7 @@ class AttackGraphNode:
     is_necessary: bool = True
     compromised_by: list[Attacker] = field(default_factory=list)
     mitre_info: Optional[str] = None
-    tags: list[str] = field(default_factory=lambda: [])
+    tags: set[str] = field(default_factory=lambda: {})
     attributes: Optional[dict] = None
```
nkakouros, line 29: The `info` lines are missing from attack steps. I can see why someone would not like them to be part of an attack graph's nodes, but these are completely inaccessible currently from what it seems. There is the `extras` dict that can be used but it is not used (and it does not seem the best place as `extras` sounds like per-application-and-use-case custom info). The simplest solution is to add an `info` attribute that is set to the language graph's step description.

maltoolbox/language/languagegraph.py
```diff
@@ -26,44 +26,64 @@
...
 class LanguageGraphAsset:
-    name: Optional[str] = None
-    associations: list[LanguageGraphAssociation] = field(default_factory=lambda: [])
-    attack_steps: list[LanguageGraphAttackStep] = field(default_factory=lambda: [])
+    name: str
+    associations: dict[str, LanguageGraphAssociation] = \
+        field(default_factory=lambda: {})
+    attack_steps: dict[str, LanguageGraphAttackStep] = \
+        field(default_factory = lambda: {})
     description: dict = field(default_factory=lambda: {})
```
nkakouros, line 44: Why call this attribute description when it is `info` fields we are copying from the MAL spec? Currently, there are three names for this information: `info` in the MAL spec, `meta` in the langspec.json and `description` here.

maltoolbox/language/languagegraph.py
```diff
@@ -1119,11 +1660,11 @@ def get_asset_by_name(
         The asset matching the name.
         None if there is no match.
         """
-        for asset in self.assets:
-            if asset.name == asset_name:
-                return asset
+        if asset_name in self.assets:
+            return self.assets[asset_name]
+        else:
+            return None
```
nkakouros, line 1666: This is not needed, it is basically a dict key lookup that you want to do. If you need to default to None (which I don't think should be the case, it should be the user's responsibility to check if the key/asset they are asking for is valid), you can use a `defaultdict`.

maltoolbox/language/languagegraph.py
```diff
@@ -888,74 +1390,149 @@ def _generate_graph(self) -> None:
...
                     ttc = attack_step_attribs['ttc'],
+                    overrides = attack_step_attribs['reaches']['overrides'] \
+                        if attack_step_attribs['reaches'] else False,
                     children = {},
                     parents = {},
-                    description = attack_step_attribs['meta']
+                    description = attack_step_attribs['meta'],
+                    tags = set(attack_step_attribs['tags'])
                 )
                 attack_step_node.attributes = attack_step_attribs
```
nkakouros, line 1400: I think this should be a private property as it is used only here to help with the separate loops and is not meant as public interface of the attack step node.

maltoolbox/model.py
```diff
@@ -808,7 +809,8 @@ def _from_dict(
         for assoc_entry in serialized_object.get('associations', []):
             assoc = list(assoc_entry.keys())[0]
             assoc_fields = assoc_entry[assoc]
```
nkakouros, line 811:
```suggestion
            assoc, assoc_fields = list(assoc_entry.items()).pop()
```

andrewbwm:
> Creating PR to be able to discuss the code. I am not the author of the code changes.

Just so that we're clear what we're discussing here. This branch is very much a work in progress. It was stopped mid-commit basically.

The only reason it got pushed to remote was in case something happened to either me or my laptop while it is on hold.

Which takes us to the second point. This is on hold until we're done with the current deadline. Once that is done I intend to pick it back up as soon as possible before I forget everything that was done with it.

maltoolbox/attackgraph/node.py
andrewbwm: Yes, this is one of the things I was having a look at when I was working on the latest commit. I think my version had some info strings in there, but some were still missing. Same thing with the tags, but I do not exactly recall what state it was left in.

nkakouros:
> Just so that we're clear what we're discussing here. This branch is very much a work in progress. It was stopped mid-commit basically.

Yes, I understand this. I just wanted to discuss some things that I saw while reading the code for when this branch is revisited.

maltoolbox/language/languagegraph.py
andrewbwm: No on both counts, I think these were all leftovers from when this function would assemble the inherited steps as well.

maltoolbox/language/languagegraph.py
andrewbwm: The idea is to have the asset only track the associations that are directly specified on it and not the inherited ones as well. And this method should return all of the associations including the inherited ones.

maltoolbox/language/languagegraph.py
andrewbwm: This makes sense to me. The `meta` name will have to be there for a bit since it's what the current MAL compiler calls it. We may deprecate later and just have converters or legacy compatibility loaders.

But what is it that we want to call this field? The `meta` name is the worst from my point of view. They all feel a bit generic, but maybe we just want to stick to `info` since that's the current MAL specification?

nkakouros: I would go with info so that there is a single name throughout the pipeline for this.

maltoolbox/language/languagegraph.py
```diff
@@ -1145,21 +1664,21 @@ def get_association_by_fields_and_assets(
...
             raise LookupError(
                 f'Failed to find asset with name \"{first_asset_name}\" in '
                 'the language graph.'
             )
 
-        second_asset = self.get_asset_by_name(second_asset_name)
+        second_asset = self.assets[second_asset_name]
         if second_asset is None:
             raise LookupError(
                 f'Failed to find asset with name \"{second_asset_name}\" in '
```
nkakouros, line 1677: The discussion here is not specific to this line. It is for wherever `self.get_asset_by_name` was removed.

#### First issue:

Is there a case now where `self.assets[key]` will return `None`? Aren't assets added to the self.assets dict when they are ready objects?

#### Exception handling:

Also, from python's side, `LookupError` is a base class. A child class called `KeyError` is what should be used instead for non-existent key lookup errors (and IndexError for non-existent list keys). The code could look like:

```
try:
    second_asset = self.assets[second_asset_name]
except KeyError as e:
    raise LanggraphError("Failed to find asset with name %s in the langgraph...", second_asset_name) from e
```

In the snippet, notice:

- the use of the %s formatter instead of f-strings
- the use of a custom exception class

If we don't want to use the custom exception class, then it is fine to not handle the exception; there is no benefit in capturing the exception just to raise it again.

#### Utility of exception handling

But there is also the question, do we expect this error to happen? Do we expect `second_asset_name` to not exist in `self.assets`? This is the language graph, so if this happens, this is a bug in the mal file...........

#### But before we discuss this more:

This function is called only by translators and ingestors. I suggest that for the rewrite we drop support for those. The neo4j ingestor uses a library that is EOL anyway. Securicad model files can be translated to json (I think). So, these two modules are not strictly needed. When the rewrite is done, we can see how we support these if we still need them and in the mean time the lang graph will have a more focused scope. For example, this function does not handle properly cases where two associations share the same field and asset names. Instead of fixing this, let's focus on the needs of the core toolbox and simulator and I am sure ingestors and translators will be able to be supported without issues and further custom additions like this function.

andrewbwm: > The discussion here is not specific to this line. It is for wherever `self.get_asset_by_name` was removed.
> #### First issue:
> 
> Is there a case now where `self.assets[key]` will return `None`? Aren't assets added to the self.assets dict when they are ready objects?

I don't think so and I also think that I've been doing too much error handling for some of these methods.

> #### Exception handling:
> 
> Also, from python's side, `LookupError` is a base class. A child class called `KeyError` is what should be used instead for non-existent key lookup errors (and IndexError for non-existent list keys). The code could look like:
> 
> ```
> try:
>     second_asset = self.assets[second_asset_name]
> except KeyError as e:
>     raise LanggraphError("Failed to find asset with name %s in the langgraph...", second_asset_name) from e
> ```
> 
> In the snippet, notice:
> 
>     * the use of the %s formatter instead of f-strings
> 
>     * the use of a custom exception class
> 
> 
> If we don't want to use the custom exception class, then it is fine to not handle the exception; there is no benefit in capturing the exception just to raise it again.
> #### Utility of exception handling
> 
> But there is also the question, do we expect this error to happen? Do we expect `second_asset_name` to not exist in `self.assets`? This is the language graph, so if this happens, this is a bug in the mal file...........

I agree with you that I think we only want to do extra exception handling if we can add anything of values to the debugging of the exception. I am not sure that it particularly helps all that much in this case. You basically know that a key is missing from the list of assets even with the default behaviour.


> #### But before we discuss this more:
> 
> This function is called only by translators and ingestors. I suggest that for the rewrite we drop support for those. The neo4j ingestor uses a library that is EOL anyway. Securicad model files can be translated to json (I think). So, these two modules are not strictly needed. When the rewrite is done, we can see how we support these if we still need them and in the mean time the lang graph will have a more focused scope.

This is something the rewrite does impact, it uses the association full names to identify them, which is `<association name>_<left_field_name>_<left_field_asset_name>_<right_field_name>_<right_field_asset_name>`. And this should always be a unique identifier so we can use this more reliably down the line. I supposed you technically could have an asset with that exact name, but even if the compiler doesn't reject it we should reject it in the mal-toolbox.

> For example, this function does not handle properly cases where two associations share the same field and asset names.

Nothing could handle it based on field names and asset names alone.

> Instead of fixing this, let's focus on the needs of the core toolbox and simulator and I am sure ingestors and translators will be able to be supported without issues and further custom additions like this function.

The ingestors and translators are out of date right now anyway. So, the focus has not been on them, it is simply that I did not check where all of these functions are used. I mostly focused on the behaviour I wanted so far, but I do think it is a useful component of the upcoming review.

maltoolbox/language/languagegraph.py
```diff
@@ -256,14 +338,23 @@ class LanguageGraphAttackStep:
     type: str
     asset: LanguageGraphAsset
     ttc: dict = field(default_factory = lambda: {})
+    overrides: bool = False
     children: dict = field(default_factory = lambda: {})
     parents: dict = field(default_factory = lambda: {})
-    description: dict = field(default_factory = lambda: {})
-    attributes: Optional[dict] = None
+    info: dict = field(default_factory = lambda: {})
+    inherits: Optional[LanguageGraphAttackStep] = None
```
nkakouros, line 345: This contradicts lines 45-47 that say:

```
    # MAL languages currently do not support multiple inheritance, but this is
    # futureproofing at its most hopeful.
    super_assets: list[LanguageGraphAsset] = field(default_factory=lambda: [])
```

Super assets is a list to allow for multiple asset parents, but step.inherits points only to one of them (the first and only). Given that we don't have semantics for multiple inheritance in MAL, trying to solve this here is futile, but the contradiction in the code remains. I suggest that any attempt to cater for multiple inheritance is dropped from the code, i.e. `super_assets` becomes `super_asset`.

maltoolbox/language/languagegraph.py
```diff
@@ -256,14 +338,23 @@ class LanguageGraphAttackStep:
...
+    tags: set = field(default_factory = lambda: {})
+    _attributes: Optional[dict] = None
```
nkakouros, line 347: Why do you keep reachable steps and requirements in `_attributes` and not directly as LanguageGraphAttackStep attributes, i.e. `step.reaches` and `step.requires`?

maltoolbox/language/languagegraph.py
```diff
@@ -294,8 +388,25 @@ def to_dict(self) -> dict:
                 else:
                     node_dict['parents'][parent].append(None)
 
+        if hasattr(self, 'requirements'):
+            node_dict['requirements'] = []
+            for requirement in self.requirements:
+                node_dict['requirements'].append(requirement.to_dict())
+
```
nkakouros, line 395: I think the attribute should be names `requires` to keep the convention that other attributes established (e.g. `inherits`, `overrides`) as well as conform to what is in other parts of the code like what gets returned from the compiler that calls this `requires`.

maltoolbox/language/languagegraph.py
```diff
@@ -411,39 +668,241 @@ def from_mar_archive(cls, mar_archive: str) -> LanguageGraph:
...
+                        attack_step.type == 'notExist':
+                    if 'requirements' in attack_step_dict:
+                        dep_chains = attack_step_dict['requirements']
+                        attack_step.requirements = []
+                        for dep_chain_dict in dep_chains:
+                            dep_chain = DependencyChain._from_dict(
+                                dep_chain_dict,
+                                lang_graph
+                            )
+                            attack_step.requirements.append(dep_chain)
```
nkakouros, line 902: This is not the best result, as it will create a step object that contains all its requirements ready to go with inheritance resolved, while the same step, if not loaded from a file but constructed on the fly from the mal and model files will not. How often is there the need to load an attack graph from a file? Why is this needed instead of constructing the attack graph always? Is it due to performance?

maltoolbox/attackgraph/analyzers/apriori.py
nkakouros, line 1: Should we add Instant, next to Enabled and Disabled?

andrewbwm: > Should we add Instant, next to Enabled and Disabled?

It's debatable, but my plan was to add all of the leftovers of #90 that do not deal with the language graph logic building anyway after I rebase that branch on the final version of this branch.

Does that seem reasonable to you?

nkakouros: Sounds good.

andrewbwm:
The latest tests use https://github.com/mal-lang/testLang ea6d872683e9193c2e2f2ef8241f0a9aba00d330 for reference.

maltoolbox/language/languagegraph.py
andrewbwm: This is probably a good idea in hindsight. A lot of the code is multiple inheritance compatible, but if we do decide to implement multiple inheritance the more complex parts would need to be discussed. For example, how do you handle the situation where multiple parents have the same attack step name. And until we have an answer to these questions it's not really helpful to just have the code be compatible for future features that might not even come. Even though I still think multiple inheritance might be desirable.

maltoolbox/attackgraph/analyzers/apriori.py
nkakouros, line 1: You have not included `Instant` in the list of "shortcircuit" distributions.

andrewbwm: No, you're right, and as per our disucssion today we should have a look at values such as Zero and Infinity.

andrewbwm: Added Instant to that list for now and added a comment that we should just apply that reasoning to all distributions that have a Bernoulli component.

maltoolbox/language/languagegraph.py
nkakouros: What I had in mind was that `self.associations` would return all associations, inherited or not. And some other property would hold the non-inherited ones. But given the design choice you have made my comment is obsolete.

maltoolbox/attackgraph/node.py
nkakouros: Should we add an `info` dict here? Or `meta`? That returns the info field from the language graph steps? Do you want me to come up with sth?

maltoolbox/language/languagegraph.py
andrewbwm: It might be a good idea to kind of get a mix of the two. We may want to cache the results of the `get_all_X` methods in a variable. And it might also be better to have the list with the inherited associations/variables as the association/variables.

maltoolbox/attackgraph/node.py
andrewbwm: Info, we really should avoid meaningless names such as "meta" as much as possible. Info is pretty bad, but you can at least get a feel for it. I will add the info strings before finalizing the merge.

nkakouros: Here we can reference the info fields from the langgraph asset's step instead of copying the strings using @property:

```
@property
def info(self):
    return self._lgstep.info
```
as an example, where _lgstep is a reference to the langgraph attack step that the current step is an instance of.

maltoolbox/language/languagegraph.py
```diff
@@ -26,44 +26,58 @@
...
-            'associations': [],
-            'attack_steps': [],
-            'description': self.description,
-            'super_assets': [],
-            'sub_assets': []
+            'associations': {},
+            'attack_steps': {},
+            'info': self.info,
+            'super_asset': self.super_asset.name if self.super_asset else "",
+            'sub_asset': self.sub_asset.name if self.sub_asset else "",
```
nkakouros, line 58:
```suggestion
            'super_asset': self.super_asset.name or "",
            'sub_asset': self.sub_asset.name or "",
```

maltoolbox/attackgraph/attackgraph.py
```diff
@@ -332,8 +172,8 @@ def _from_dict(
                 'is_necessary' in node_dict else True
             ag_node.mitre_info = str(node_dict['mitre_info']) if \
                 'mitre_info' in node_dict else None
-            ag_node.tags = node_dict['tags'] if \
-                'tags' in node_dict else []
+            ag_node.tags = set(node_dict['tags']) if \
+                'tags' in node_dict else set()
```
nkakouros, line 176:
```suggestion
            ag_node.tags = set(node_dict.get('tags', []))
```

Would it be better to return a set from the compiler already? I think it would. In general, there are such checks scattered all over the place in the app code and we could keep it less noisy if the compiler did such things. I have a TA that started working today on the compiler. I can ask him to review the toolbox code for such things and make a list of things that we already do in the compiler.

maltoolbox/attackgraph/attackgraph.py
```diff
@@ -493,6 +333,176 @@ def attach_attackers(self) -> None:
...
+                          should apply to. Initially it will contain the
+                          asset to which the attack step belongs
+        expr_chain      - the expressions chain we are following
+
+        Return:
+        A list of all of the target assets.
+        """
+
+        if expr_chain is None:
+            # This is an attack step targeting the same assets
```
nkakouros, line 359: I think it would be nice to have an example of how this would look like in mal code in each of the cases. For instance, here, it is hard for me to visualize what you are saying in the code. I am sure it is simple, but i just can't get my head around it.

nkakouros: I get it now. You pass all children of an attack step through this method here and this conditional applies to a simple step that is reachable, ie:

```
| a_step
  -> another_step
```

nkakouros: Perhaps:

```suggestion
            # This is an attack step in the same asset as the parent:
            # | parent_step
            #   -> this_step
```

maltoolbox/attackgraph/attackgraph.py
```diff
@@ -493,6 +333,176 @@ def attach_attackers(self) -> None:
...
+        match (expr_chain.type):
+            case 'union' | 'intersection' | 'difference':
+                # The set operators are used to combine the left hand and
+                # right hand targets accordingly.
+                if not expr_chain.left_link:
+                    raise LanguageGraphException('"%s" step expression chain'
+                        ' is missing the left link.' % expr_chain.type)
+                if not expr_chain.right_link:
+                    raise LanguageGraphException('"%s" step expression chain'
+                        ' is missing the right link.' % expr_chain.type)
```
nkakouros, line 378: If this is the case and there is indeed such an error, is the underlying MAL code valid? If not, this should be checked for at the compiler level.

maltoolbox/language/languagegraph.py
```diff
@@ -26,44 +26,58 @@
 
 logger = logging.getLogger(__name__)
 
+def get_asset_name_from_full_attack_step(
+        attack_step_full_name: str) -> str:
+    return attack_step_full_name.split(':')[0]
+
+def get_attack_step_name_from_full_attack_step(
+        attack_step_full_name: str) -> str:
+    return attack_step_full_name.split(':')[1]
```
nkakouros, line 35: These two are always used in pair. I suggest:

```suggestion
def decompose_full_name(
        full_name: str) -> list[str]:
    return attack_step_full_name.split(':')
```

And use it like:

```
target_asset_name, target_attack_step_name = decompose_attack_step_full_name(target_full_attack_step_name)
```

maltoolbox/attackgraph/attackgraph.py
```diff
@@ -493,6 +333,176 @@ def attach_attackers(self) -> None:
...
+                )
+                rh_targets = self._follow_expr_chain(
+                    model,
+                    target_assets,
+                    expr_chain.right_link
+                )
+
+                match (expr_chain.type):
+                    # Once the assets become hashable set operations should be
+                    # used instead.
```
nkakouros, line 392: Should this be a TODO?

```suggestion
                    # TODO: Once the assets become hashable set operations
                    # should be used instead.
```

maltoolbox/attackgraph/attackgraph.py
```diff
@@ -493,6 +333,176 @@ def attach_attackers(self) -> None:
 
             attacker.entry_points = list(attacker.reached_attack_steps)
 
+    def _follow_expr_chain(
+            self,
+            model: Model,
+            target_assets: list[Any],
```
nkakouros, line 339: Is this really `list[Any]`? Isn't it always:

```suggestion
            target_assets: list[LanguageGraphAsset],
```
?

maltoolbox/attackgraph/attackgraph.py
```diff
@@ -493,6 +333,176 @@ def attach_attackers(self) -> None:
...
+        target_assets   - the list of assets that this expressions chain
+                          should apply to. Initially it will contain the
+                          asset to which the attack step belongs
+        expr_chain      - the expressions chain we are following
+
+        Return:
+        A list of all of the target assets.
+        """
+
+        if expr_chain is None:
```
nkakouros, line 358: Does this mean that the type annotation a few lines above should change to `Optional[ExpressionsChain]`?

maltoolbox/attackgraph/attackgraph.py
nkakouros: I also think that this should be implemented before merging this PR.

maltoolbox/language/languagegraph.py
nkakouros: If it is None, won't the check `if self.sub_asset` fail and `else` will happen? Which is the same with `self.sub_asset.name or ""` where `self.sub_asset.name` will be falsy and the second element in the `or` expression will be chosen.

maltoolbox/attackgraph/attackgraph.py
```diff
@@ -493,6 +333,176 @@ def attach_attackers(self) -> None:
...
+                # associated assets given the specified field name.
+                if not expr_chain.fieldname:
+                    raise LanguageGraphException('"field" step expression '
+                        'chain is missing fieldname.')
+                new_target_assets = []
+                for target_asset in target_assets:
+                    new_target_assets.extend(model.\
+                        get_associated_assets_by_field_name(target_asset,
+                            expr_chain.fieldname))
+                return new_target_assets
```
nkakouros, line 424: The `get_associated_assets_by_field_name` functions is this:

https://github.com/mal-lang/mal-toolbox/blob/681e6c64af1234c3cdab75ceb94be3455ee4162d/maltoolbox/model.py#L598-L636

The `target_asset` here is a `LanguageGraphAsset` right? This is against the signature of that method. 

Also, I think that method could be simplified:

```python
def get_associated_assets_by_field_name(
        self,
        asset: SchemaGeneratedClass,
        field_name: str
    ) -> list[SchemaGeneratedClass]:
    """
    Get a list of associated assets for an asset given a field name.

    Arguments:
    asset           - the asset whose fields we are interested in
    field_name      - the field name we are looking for

    Return:
    A list of assets associated with the asset given that match the
    field_name.
    """
    logger.debug(
        'Get associated assets for asset "%s"(%d) by field name %s.',
        asset.name, asset.id, field_name
    )

    return [
        associated_asset
        for association in asset.associations
        if (associated_asset := getattr(association, field_name))
    ]
```

maltoolbox/attackgraph/attackgraph.py
```diff
@@ -493,6 +333,176 @@ def attach_attackers(self) -> None:
...
+                    additional_assets = self._follow_expr_chain(
+                        model, new_target_assets, expr_chain.sub_link)
+
+                    new_target_assets = []
+                    for additional_asset in additional_assets:
+                        if additional_asset not in accumulated_target_assets:
+                            accumulated_target_assets.append(additional_asset)
+                            new_target_assets.append(additional_asset)
+
+                return accumulated_target_assets
```
nkakouros, line 442: I found this block a bit confusing.

```suggestion
                accumulated_target_assets = [target_assets]
                for asset_batch in accumulated_target_assets:
                    additional_assets = self._follow_expr_chain(
                        model, asset_batch, expr_chain.sub_link)

                    # The need for new_assets will be gone once things are sets
                    new_assets = []
                    for additional_asset in additional_assets:
                        if additional_asset not in asset_batch:
                            new_assets.append(additional_asset)
                     
                     if new_assets:
                         accumulated_target_assets.append(new_assets)

                return [asset for batch in accumulated_target_assets for asset in batch]
```

maltoolbox/attackgraph/attackgraph.py
```diff
@@ -493,6 +333,176 @@ def attach_attackers(self) -> None:
...
+                        ' is missing the right link.')
+                lh_targets = self._follow_expr_chain(
+                    model,
+                    target_assets,
+                    expr_chain.left_link
+                )
+                rh_targets = self._follow_expr_chain(
+                    model,
+                    lh_targets,
+                    expr_chain.right_link
```
nkakouros, line 490: Will this be sth else than None? If yes, I have not understood what is going on, no need to explain. If no, maybe have it explicitly say None?

```suggestion
                    None,
```

maltoolbox/attackgraph/attackgraph.py
```diff
@@ -514,57 +524,70 @@ def _generate_graph(self) -> None:
...
+            lang_graph_asset = self.lang_graph.assets[asset.type]
+            if lang_graph_asset is None:
+                raise LookupError(
+                    f'Failed to find asset with name \"{asset.type}\" in '
+                    'the language graph.'
+                )
 
-            for attack_step_name, attack_step_attribs in attack_steps.items():
+            for attack_step_name, attack_step in \
+                    lang_graph_asset.attack_steps.items():
```
nkakouros, line 535: Is `attack_step_name` used? I think you use `attack_step.name` everywhere, in which case you can loop through the values only:

```suggestion
            for attack_step in lang_graph_asset.attack_steps.values():
```

maltoolbox/attackgraph/attackgraph.py
```diff
@@ -514,57 +524,70 @@ def _generate_graph(self) -> None:
...
+                            # exist in the model.
+                            if target_assets:
+                                existence_status = True
+                                break
+
+                    case _:
+                        pass
+
+                mitre_info = attack_step.info['mitre'] if\
+                    'mitre' in attack_step.info else None
```
nkakouros, line 575:
```suggestion
                mitre_info = attack_step.info.get('mitre', None)
```

maltoolbox/attackgraph/attackgraph.py
```diff
@@ -576,42 +599,67 @@ def _generate_graph(self) -> None:
...
+                    'asset link')
+            lang_graph_asset = self.lang_graph.assets[ag_node.asset.type]
+
+            lang_graph_attack_step = lang_graph_asset.attack_steps[\
+                ag_node.name]
+
+            while lang_graph_attack_step:
+                for child in lang_graph_attack_step.children:
+                    for target_attack_step, expr_chain in \
+                            lang_graph_attack_step.children[child]:
```
nkakouros, line 614:
```suggestion
                for child in lang_graph_attack_step.children.values():
                    for target_attack_step, expr_chain in child:
```

maltoolbox/attackgraph/attackgraph.py
```diff
@@ -576,42 +599,67 @@ def _generate_graph(self) -> None:
...
+                                    )
+                                    raise AttackGraphStepExpressionError(
+                                        msg % (
+                                            target_node_full_name,
+                                            ag_node.full_name,
+                                            ag_node.id
+                                        )
+                                    )
+
+                                assert ag_node.id is not None
```
nkakouros, line 645: This seems a bit out of place here. `ag_node` is not altered in this loop at all, right?

maltoolbox/attackgraph/attackgraph.py
```diff
@@ -576,42 +599,67 @@ def _generate_graph(self) -> None:
...
+                                assert target_node.id is not None
```
nkakouros, line 646: This seems a bit out of place here. `ag_node` is not altered in this loop at all, right? 

Also, is there ever a case where a node may not have an id? If it is a general requirement, I think we should enforce it somewhere else and not check for it here. Maybe, we could do sth in the node class like:

```python
@property
def id(self):
    if self._id is None: ...fail...
    return self._id
```

maltoolbox/attackgraph/attackgraph.py
```diff
@@ -576,42 +599,67 @@ def _generate_graph(self) -> None:
...
+                                        ag_node.id,
+                                        target_node.full_name,
+                                        target_node.id
+                                    )
+                                )
+                                ag_node.children.append(target_node)
+                                target_node.parents.append(ag_node)
+                if bool(lang_graph_attack_step.overrides):
+                    break
+                lang_graph_attack_step = lang_graph_attack_step.inherits
```
nkakouros, line 661: Why do you need to cast to `bool`? Isn't it already a bool? 

```suggestion
                if not lang_graph_attack_step.overrides:
                    lang_graph_attack_step = lang_graph_attack_step.inherits
```

maltoolbox/attackgraph/attackgraph.py
nkakouros: This is in `from_dict` I notice now, so it is an invalid comment.

maltoolbox/attackgraph/attackgraph.py
andrewbwm: Isn't this the transitive block? It needs to keep running until it no longer finds new assets and I don't like having for loops that vary in size over iteration.

maltoolbox/language/languagegraph.py
andrewbwm: Just for the record, we ended up with `@cached_property` (this is necessary, cache and property cache the property between language graphs that have the same assets if another one is loaded. This made tests fail and in general is not what we want). And it doesn't really work for variables as they kind of get build when fetching them the first few times around. I think that's fine, but if we spot performance issues with it we can figure out an optimization for variables specifically.

maltoolbox/language/languagegraph.py
andrewbwm: Let me clarify this, `self.sub_asset` is missing, not `self.sub_asset.name`.

maltoolbox/attackgraph/attackgraph.py
nkakouros: It was just a suggestion with what seemed to me easier to follow code. Ignore this if you think it adds nothing.

maltoolbox/language/languagegraph.py
nkakouros: ah, I see, I didn't read that code carefully.

maltoolbox/attackgraph/attackgraph.py
andrewbwm: This is not at all a criticism, but it should be blatantly obvious by now that you can't ask my brain to not explain something. It's just not how it works.

Your overall gut feeling is correct, but it's what led to my initial wrong assumption. Yes, most often the right side of a collect will be None. However, you can have stuff like (exprA).(exprB), exprB would have to contain the attack step somewhere in there. So that resulting expression would have exprA in the left link and exprB in the right link, and exprB has to contain at least one other collect expression that has the attack step on the right hand side which would be a None right link.

Did I "not explain" that clearly enough?

nkakouros: Ah right, I see it now. Thanks for clarifying !

andrewbwm: Just for future reference in case I/we need it. You don't actually need the brackets for exprA, any expression that has exprA.(exprB) has a non-None right link because the right link will always be a collect that eventually has to contain a None on a right link.

maltoolbox/attackgraph/attackgraph.py
andrewbwm: These are instance model assets, not language graph asset types.

nkakouros: Aha, so this will have to change if we remove the class factory and instead of one class per model asset we have instances of a single ModelAsset class.

maltoolbox/attackgraph/attackgraph.py
```diff
@@ -493,6 +333,176 @@ def attach_attackers(self) -> None:
 
             attacker.entry_points = list(attacker.reached_attack_steps)
 
+    def _follow_expr_chain(
```
nkakouros, line 336: I think this method should be cached. There can be some quite expensive expressions to evaluate that may repeat in various other expressions. There is an issue with `@cache` that expects the method args to be hashable but lists and Model.assets are not. We can replace lists with tuples and the first issue is solved. About the second issue, once the class factory is gone this will be possible for us to tackle. Should we defer caching to the factory removal PR or try to tackle here?

andrewbwm: Let's defer it for now. This pull request is overly complicated as is.

nkakouros: https://github.com/mal-lang/mal-toolbox/issues/96

maltoolbox/attackgraph/attackgraph.py
andrewbwm: These were added because mypy would complain about it.

maltoolbox/attackgraph/attackgraph.py
```diff
@@ -493,6 +345,178 @@ def attach_attackers(self) -> None:
 
             attacker.entry_points = list(attacker.reached_attack_steps)
 
+    @cache
```
nkakouros, line 348: Actually, I noticed that the use of `@cache` actually slows things down. For a language I used, the hit rate for the cache was ~36%, or 1/3 of the time. This low and the cost of maintaining the cache outperforms its value. Sure, this depends on the mal spec used, but I think having a low cache hit rate should probably be the case for MAL specs in general; variable definitions are contained in assets and it does not make sense for many assets to share similar variable definitions.

The performance overhead of maintaining the cache was negligible, about 2 seconds more on a ~40s total runtime. As a percentage it is a lot, but I don't think it matters. But to keep the code simple maybe we want to remove it anyway? If we do remove it, there is no need to use frozensets for example, we can use just `sets`. This will also save some milliseconds from casting frozensets to sets in a couple of lines in this method.

What do you say?

andrewbwm: I like keeping the code simple and clean because it's easier to debug and profile. So I'm definitely for switching to regular sets and removing caching where it isn't needed.

maltoolbox/attackgraph/attackgraph.py
andrewbwm: It should change and I will clarify the previous point as well with a comment in the code.

maltoolbox/language/languagegraph.py
nkakouros: I think it looks fine as is, maybe in some future rewrite (!) there is a better idea on how to do this.

maltoolbox/language/languagegraph.py
nkakouros: Yes, it does. Let's leave as is.

maltoolbox/attackgraph/attackgraph.py
andrewbwm: It is not a `LanguageGraphAsset` it is an instance of a class in the language classes factory. So, the signature is correct, but you're right that that the `get_associated_asset_by_fieldname` method could have been simplified a lot. That commit will be pushed momentarily.

maltoolbox/attackgraph/attackgraph.py
andrewbwm: It is used for the name field and the debug log.

nkakouros: we can use attack_step.name instead right?

maltoolbox/attackgraph/attackgraph.py
andrewbwm: I think now that we have all of the info fields exposed via the language graph attack step we shouldn't handle the `mitre_info` separetely anymore. Do you agree?

maltoolbox/attackgraph/attackgraph.py
nkakouros: It adds one more variable to keep track of which is not needed at all. You have to keep one more variable in your mind to read the code. Also, it creates confusion, doesn't the step object already have the name, why carry it around separately, is it different to step.name, etc.

maltoolbox/attackgraph/attackgraph.py
nkakouros: I think it makes sense.

maltoolbox/attackgraph/attackgraph.py
andrewbwm: It's this weird thing where the id is a property of the `AttackGraphNode`, but it is determined when the node is added to the `AttackGraph`. So, there is a bit of a period where the `AttackGraphNode` exists, but it does not have its id determined.

I think these asserts in particular were here because mypy would complain that they might be None since they're optional. Your suggestion would also fix that, but I think the current asserts are fine too. Having a wrapper around such a simple property feels a bit confusing, not that the current implementation isn't confusing where the id is sort of semi-mandatory, it has a very narrow window where it isn't required.

maltoolbox/attackgraph/attackgraph.py
andrewbwm: The comment is correct, it is already a bool, it doesn't need the cast. The code suggestion is not correct, this is a while loop that iterates through the language graph attack steps inheritance hierarchy.

If it ever runs into an overriden step it needs to exit that loop. Your suggestion doesn't do that which leads to an infinite loop if it does have an overriden step because the `lang_graph_attack_step` will no longer change and the loop does not break.

PR #88, opened 2024-10-16
Fix parent step targets being overwritten by child step targets

apelG:
In the else-case of _get_attacks_for_asset_type, i.e. when a child asset has an attack step that

1. Already exists on parent
2. Does not override parent
3. Has ["reaches"]["stepExpressions"] (i.e., points somewhere)

the code will override the *parent's* attack step with the targets of the child. I.e.:

```
Parent
| attackStep
 -> as1

Child
| attackStep
 +> as2,
      as3
```

Becomes:

``` 
Parent
| attackStep
  -> as2,
      as3
```

This will also cause a crash if as2, as3.. does not exist on `Parent`.

The fix can technically be applied inside the else-clause just before the `extend()`-call, but I put the fix before the return to catch any other edge-cases that show up (either now or in the future).


andrewbwm:
I can have a look and see if this is still an issue after we finish the language graph rework is complete. If you want this merged sooner because you need it before that integration is finished(most likely around the end of November) let me know.

Same thing with the other branch you pushed during the summer, some of those fixes I think will be solved the by the language graph rework, but I will need to look into it when it's done to figure out if there are some that were not addressed by the language graph redesign.

apelG:
Is there a downside to merging the fix now? Even if the rework fixes the issue, you say that it is >1 month before it will be complete. Considering that this bug can cause both crashes and (perhaps more importantly) outright incorrect behavior, it seems iffy to leave it around. Especially since the fix is just a 1-line change that won't affect anything else in the module.

As for the other changes I pushed during the summer, I think it would be great if we can go through the rework together once its done so that we can verify that the logic issues that @vwpi found have been addressed.

Also, just out of curiosity, what is the language graph rework supposed to accomplish? Is it just a code refactor, or something more?

andrewbwm:
> Is there a downside to merging the fix now? Even if the rework fixes the issue, you say that it is >1 month before it will be complete. Considering that this bug can cause both crashes and (perhaps more importantly) outright incorrect behavior, it seems iffy to leave it around. Especially since the fix is just a 1-line change that won't affect anything else in the module.
> 
> As for the other changes I pushed during the summer, I think it would be great if we can go through the rework together once its done so that we can verify that the logic issues that @vwpi found have been addressed.
> 
> Also, just out of curiosity, what is the language graph rework supposed to accomplish? Is it just a code refactor, or something more?

I will merge this now and I'll get back to you when the language graph rework is done. It is a pretty big reworking, so not a refactor, but I don't have the time to get into it right now. Once it is done we can have a look and see which bugs are left over.

andrewbwm:
> Hm, interesting. I thought it just required one level of inheritance to happen, but maybe the trigger condition is even more complicated than I initially thought. For reference, this is the original structure that crashed when I tried to create the graph:
> 
> ```
> EmptyParent
> (doesn't define attackstep)
> 
> Child1 extends EmptyParent
> | attackstep
> 
> Child2 extends Child1
> (doesn't change attackstep)
> 
> Child3 extends Child2
> | attackstep
>   +> target1
> 
> Child4 extends Child3
> (doesn't change attackstep)
> 
> FinalChild extends Child4
> | attackstep
>   +> target1,
>        target2,
>        target3,
>        target4,
> ```
> 
> This results in a langgraph with:
> 
> ```
> Child3
> | attackstep
>   +> target1,
>        target2,
>        target3,
>        target4
> ```
> 
> Which is ofc incorrect, and which results in a crash since target2-4 are not defined for Parent3.
> 
> I'll try to dig around a bit and see if I can figure out what the exact condition is.

So, this is definitely something that the language graph rework would impact, because I changed how inheritance works for attack steps. Now `Child3:attackStep` for example would say that it inherits `Child2:attackStep` and it also contains `target1`. So, inheritance is maintained in the hierachy of the Language Graph, it isn't just copied over as it was before.

The question that I have is if this is something you need patched right now because you're using it? If you are, I can just apply the original one line commit after a bit more testing to make sure it doesn't mess up any of our existing language graphs, which I assume it does not. If it isn't urgent I'd rather just wait for the reworked language graph.

apelG:
Ok, sounds good. I have already fixed it locally so for my personal use there is no need to merge it to main right now.

PR #90, opened 2024-12-02
Step expression fixes

nkakouros:
I didn't spent time optimizing any of them as they will soon be obsolete.

### DIFF
--- a/maltoolbox/attackgraph/attackgraph.py
+++ b/maltoolbox/attackgraph/attackgraph.py
@@ -76,23 +76,21 @@ def _process_step_expression(
             match (step_expression['type']):
                 case 'union':
                     new_target_assets = lh_targets
-                    for ag_node in rh_targets:
-                        if next((lnode for lnode in new_target_assets \
-                            if lnode.id != ag_node.id), None):
-                            new_target_assets.append(ag_node)
+                    new_target_assets.extend(rh_targets)
+                    for target in new_target_assets.copy():
+                        if new_target_assets.count(target) > 1:
+                            new_target_assets.remove(target)
 
                 case 'intersection':
-                    for ag_node in rh_targets:
-                        if next((lnode for lnode in lh_targets \
-                            if lnode.id == ag_node.id), None):
-                            new_target_assets.append(ag_node)
+                    new_target_assets = []
+                    for target in rh_targets:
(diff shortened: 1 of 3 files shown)


maltoolbox/attackgraph/attackgraph.py
```diff
@@ -126,22 +127,21 @@ def _process_step_expression(
...
-            # The transitive expression is very similar to the field
-            # expression, but it proceeds recursively until no target is
-            # found and it and it sets the new targets to the entire list
-            # of assets identified during the entire transitive recursion.
-            new_target_assets = []
+            # The transitive expression is very similar to the field expression,
+            # but it proceeds recursively until no target is found and it sets
+            # the new targets to the entire list of assets identified during the
+            # entire transitive recursion.
+            new_target_assets = target_assets
```
nkakouros, line 134: Is this line correct? Do the target_assets by any chance get returned back from the transitive expression which would have them end up being included twice?

andrewbwm: Yes, that is a potential problem, but it can only happen if an asset reference back to itself directly or indirectly. What we want is to turn this into a set because otherwise we would have to check every time for the `transitive` and field `type` step expressions.

The catch there is that we need to make assets hashable, whcih we probably can since we can just make hashes of their ids, since those are unique.

nkakouros: Yes, using the id for the hash makes sense. If an asset is referenced back, won't this cause infinite loops?

andrewbwm: Yes, we would need to have a check to see if the `new_target_assets` and `target_assets` differ before allowing further recursion. That's really only a problem for the `field` anyway, since `transitive` already has that issue, it has to know when to stop, and that is when it gets no new assets, so we basically need the check from `transitive` also apply for `field` step expressions.

nkakouros: Is it good enough as it is here? Maybe only the new lang graph goes to such lengths?

andrewbwm: But we're not merging this pull request anyway, right? This is just here to help out with updates in the new language graph integration. Or did I misunderstand its purpose?

PR #92, opened 2024-12-04
Replace classes factory with sth lightweight

nkakouros:
I replaced the json schema based classes factory with sth more lightweight. I tried to leave the interface of the factory and the classes that get generated and used as much the same as before as possible. The code seems to work, some tests fail and some things are not implemented cleanly. If the approach gets the green light, the code will be refactored to be cleaner.

If there is no requirement for the model assets to correspond to different classes based on their asset type, i.e. if we don't need to have classes like `class Application`, `class Interface`, etc, then we can drop the factory altogether and simply create instances of `ModelAsset` that have the appropriate names and types.

Also, it seems to me that the code could be simplified if we had a standard interface for what a model asset or step should contain. For instance, there is no need to check if an asset has defenses and then pull `asset.defenses`. We can set `Asset.defenses = {}` by default. 

A question to you, is there sth that the old schema code did that the one in this PR does not?

A task list for future work:

- [ ] Drop the factory altogether
- [ ] Simplify code by having a standard interface for ModelAsset and ModelStep
- [ ] Fix tests
- [ ] Fix linters

I don't know if you have these in place already, but for such refactorings, checking if imports are used by ruff would be great.


nkakouros:
I added a second commit that completely retires the `maltoolbox.language.classes_factory` module. See the first commit first, then the second.

nkakouros:
The 3rd commit uses ModelAsset and ModelAssoc directly without having to go through `type`.

nkakouros:
Btw, there is a set of `remove_*` methods that remove assets and associations. Are these used anywhere?

andrewbwm:
> Btw, there is a set of `remove_*` methods that remove assets and associations. Are these used anywhere?

Yes, https://github.com/mal-lang/mal-gui uses them.

andrewbwm:
> A question to you, is there sth that the old schema code did that the one in this PR does not?

I think there might be some things missing the in handling of defenses, especially their default values.

But after the work on the language integration is done I can have a more thorough and in-depth look at what this change would entail. My initial evaluation is that I would prefer this more lightweight approach and it suits our needs.

mrkickling:
Just starting to look into this. It is a bit hard to review before the new Lang graph is merged but thoughts so far:

About ModelAsset, ModelAssociation:
- We should make them dataclasses or at least have predefined instance variables (I saw you mentioned something about having a standard interface, all for that, otherwise we get the same issue as before when fetching instance variables)
- Associations need to verify asset types when created
- It should be easy to create an asset/association of a certain type (this might be the case already)

mrkickling:
When we refer to language graph assets in the model context we should refer to them as asset types.
Same things with language graphs associations.

maltoolbox/model.py
```diff
@@ -855,4 +821,33 @@ def load_from_file(
             serialized_model = load_dict_from_json_file(filename)
         else:
             raise ValueError('Unknown file extension, expected json/yml/yaml')
-        return cls._from_dict(serialized_model, lang_classes_factory)
+        return cls._from_dict(serialized_model, lang_graph)
+
+class ModelAsset:
```
mrkickling, line 826: Suggestion for something more standardized (it is missing some stuff like what the defenses/states are in relation to the asset)

```
@dataclass
class ModelAsset:
    """
    A ModelAsset is an instance of a LanguageGraphAsset
    which means that it contains additional data related to the
    specific asset.
    """

    name: str
    type: LanguageGraphAsset
    defenses: list = field(default_factory=lambda: [])

    # Getters/utility methods could be added

@dataclass
class ModelAssociation:
    """
    A ModelAssociation is an instance of a LanguageGraphAssociation
    which means that it contains additional data related to the
    specific association.
    """

    name: str
    type: LanguageGraphAssociation

    left_field_assets: list[ModelAsset]
    right_field_assets: list[ModelAsset]

    # Getters/utility methods could be added
    # maybe a custom __init__  to add validation
``

mrkickling: I realized your solution still used the factories, I had missed that.
I think it makes more sense that all assets and associations are just ModelAssets and ModelAssociations.
Let me know what you guys think.

nkakouros: TBH, I don't even remember what i did. I'll revisit it and reply here.

mrkickling: We need a rebase as well, but  I am not sure how to handle some of the conflicts. Would you mind checking if it is easier for you to solve the conflicts? Otherwise I will try to sort it out.

nkakouros: I'll wait for the lang graph branch to be merged and then probably create a new PR against the main branch.

PR #93, opened 2024-12-19
Add function 'propagate_viability_from_unviable_node'

mrkickling:
That returns which attack steps were made unviable because of unviable parent node.

This was previously done in MAL Simulator, but fits better in the toolbox imo.

### DIFF
--- a/maltoolbox/attackgraph/analyzers/apriori.py
+++ b/maltoolbox/attackgraph/analyzers/apriori.py
@@ -183,7 +183,6 @@ def prune_unviable_and_unnecessary_nodes(graph: AttackGraph) -> None:
 
 def propagate_viability_from_unviable_node(
         unviable_node: AttackGraphNode,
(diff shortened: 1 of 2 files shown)


maltoolbox/attackgraph/analyzers/apriori.py
```diff
@@ -173,3 +179,54 @@ def prune_unviable_and_unnecessary_nodes(graph: AttackGraph) -> None:
...
+    that are no longer viable because of it.
+
+    Propagate recursively via children as long as changes occur.
+
+    Arguments:
+    unviable_node               - the node to propagate viability from
+    attack_steps_made_unviable  - list of the attack steps that have been
+                                  made unviable by a defense enabled in the
+                                  current step. Builds up recursively.
+    """
```
andrewbwm, line 200: I don't like the fact that it has the result both as a parameter and the return value. I would have it only in the return and build it up via recursion.

andrewbwm (changes requested):

PR #95, opened 2025-01-16
Add LookupDict for faster and simpler lookups

nkakouros:
In various places, we have a list of items in the form of a dict for faster lookup. Still, it is often the case that we want to lookup items in the list (dict.values() actually) based on an attribute other than the one used as the dict keys. This PR introduces the LookupDict and in the future it will replace such simple dicts (such as attackgraph.nodes) with LookupDicts. LookupDicts work:

```
# like typical dicts
attackgraph.nodes['Network:attach']

# can filter based on  any attribute
attackgraph.nodes.lookup("id", 0)

# more complex filtering
attackgraph.nodes.lookup("id", [0,1,2], "in")
attackgraph.assets.lookup("name", "comprom", "contains")

### DIFF
--- a/maltoolbox/attackgraph/analyzers/apriori.py
+++ b/maltoolbox/attackgraph/analyzers/apriori.py
@@ -162,7 +162,7 @@ def calculate_viability_and_necessity(graph: AttackGraph) -> None:
     graph       - the attack graph for which we wish to determine the
                   viability and necessity statuses for the nodes.
     """
-    for node in graph.nodes:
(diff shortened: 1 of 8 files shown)


nkakouros:
I have applied the proposed `LookupDict` to the `nodes` property of the `AttackGraph`. The saved lines of code are already more than the implementation of the `LookupDict`. If the `LookupDict` is used in more places the savings will be greater. This gets translated to easier to read and maintain code, together with performance improvements.

PR #99, opened 2025-01-17
Lint code with ruff

nkakouros:


nkakouros:
This will wait for the lang graph branch to be merged, and the auto fixes will be applied again. But in general how does this look like? Is there sth you want to change?

I think having such a setup and configuration will make life a lot easier. For instance, we won't have to think again about formatting, the code will have a consistent look and some bugs may fixed in the mean time. 

The drawback of this PR is that despite the auto fixes there are almost 2k issues that need to be resolved.

maltoolbox/__init__.py
```diff
@@ -16,41 +15,40 @@
...
 import logging
+import os
 
 ERROR_INCORRECT_CONFIG = 1
 
-CONFIGFILE = os.path.join(
-    os.path.dirname(os.path.abspath(__file__)),
-    "default.conf"
-)
+CONFIGFILE = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'default.conf')
```
mrkickling, line 40: Line length seems too long - I think me and Andrei have line length set to 79 or 80

maltoolbox/attackgraph/analyzers/apriori.py
```diff
@@ -105,30 +113,34 @@ def evaluate_viability(node: AttackGraphNode) -> None:
...
-    """
-    Arguments:
+    """Arguments:
     graph       - the node to evaluate necessity for.
+
     """
-    match (node.type):
+    match node.type:
         case 'exist':
-            assert isinstance(node.existence_status, bool), \
```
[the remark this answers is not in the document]
nkakouros: According to chatgpt the are three main reasons:

- a space after the backslash will result in error
- it fits other recommended patterns of splitting like:

```python
items = [
    "apple",
    "banana",
    "cherry",
]
```

- it follows the python approach of using parentheses to split long lines like in this case:
```python
q = (
    "this "
    "is a single"
    " string"
)
```

I would also add that it makes the code breath more by using more spaces.

To me the backslashes look really weird and make the code not stand out as python at first glance.

maltoolbox/__init__.py
nkakouros: We can configure it. 88 is the common setting and the default in ruff. If you look at pyproject.toml there is a rationale behind it. 

About 88 being the default, Black, which is what ruff follows, [says](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length):

> Line length
> 
> You probably noticed the peculiar default line length. Black defaults to 88 characters per line, which happens to be 10% over 80. This number was found to produce significantly shorter files than sticking with 80 (the most popular), or even 79 (used by the standard library). In general, [90-ish seems like the wise choice](https://youtu.be/wf-BqAjZb8M?t=260).
> 
> If you’re paid by the lines of code you write, you can pass --line-length with a lower number. Black will try to respect that. However, sometimes it won’t be able to without breaking other rules. In those rare cases, auto-formatted code will exceed your allotted limit.

It is also the case that if you have a hard limit of 80, then you may have one or two characters that go over it. It is a good practice, and there are plugins that allow for a 10% more line length than the configured one so that you don't have to sacrifice readability to satisfy a linter or a hard rule. 10% of 80 is 88. This may be a coincidence, but I doubt it. 

In general, my approach to these things has been to go with what the defaults are, and by defaults, I mean also the tools of the trade such as ruff and black. This is similar to how I don't override application defaults anymore; use and learn what is there instead of learning your own things. This keeps peace of mind and reduces the maintenance burden. 

But if you need this changed, it can be changed.

mrkickling: I personally think 88 is too long. Looking at that line it is too long (more breath with shorter lines).

maltoolbox/attackgraph/analyzers/apriori.py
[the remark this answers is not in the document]
mrkickling: I sometimes have the same feeling with backslashes, but in some cases I now think it looks better (in this case for example). But it is all just matter of taste, and I will not put in a veto when it comes to parenthesis, it probably looks better in general.

nkakouros:
I autofixed whatever ruff could fix. I then ignored all failing rules. This can be merged as is now and the ignored rules can be uncommented one by one and the issues addressed in separate PRs.

PR #101, opened 2025-01-17
[lang-graph-integration] Uses sets for expr ops and caches result

nkakouros:
This is also for the lang-graph-branch, fixing set operations to use `set()` and enabling the caching of `_follow_expr_chain`.

### DIFF
--- a/maltoolbox/attackgraph/attackgraph.py
+++ b/maltoolbox/attackgraph/attackgraph.py
@@ -6,6 +6,8 @@ import copy
 import logging
 import json
 
+from functools import cache
+from itertools import chain
 from typing import TYPE_CHECKING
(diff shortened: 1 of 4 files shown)


nkakouros:
I have some good news and some bad news.

The bad news is that there was a bug in  how I changed the transitive step, I uses a wrong variable somewhere. The code still worked, it was just that not all the assets were collected. The tests passed. This means that the tests do not validate the transitive operation works as expected.

The good news is I fixed that and now it seems that the big model that was taking 2 hours to convert into an attackgraph is now loaded in seconds!! And the output seems to be correct!

andrewbwm:
I will still review this next monday in more detail

PR #103, opened 2025-01-23
Retire Language Classes Factory

andrewbwm:
The language classes factory is cumbersome and not really needed. Removing it will also improve performance since validators will not need to be used every time. But we will need to create restriction enforcement for associations and defences.


andrewbwm:
#92 reference to the old pull request that was attempting to merge into the #87 .

maltoolbox/model.py
```diff
@@ -786,16 +743,11 @@ def _from_dict(
...
-            # TODO: remove this when factory goes away
-            asset_type_class.__hash__ = lambda self: hash(self.name)  # type: ignore[method-assign,misc]
-
-            if asset_type_class is None:
-                raise LookupError('Failed to find asset "%s" in language'
-                ' classes factory' % asset_object['type'])
-            asset = asset_type_class(name = asset_object['name'])
+            asset = ModelAsset(
+                name=f"Asset_{asset_object['type']}",
+                lg_asset=lang_graph.assets[asset_object['type']],
```
mrkickling, line 748: What do you think about calling `lg_asset` asset_type instead?

maltoolbox/model.py
```diff
@@ -808,19 +760,21 @@ def _from_dict(
...
-                get_association_class_by_fieldnames(assoc, field1, field2)
-            if assoc_type_class is None:
-                raise LookupError('Failed to find association "%s" with '
-                    'fields "%s" and "%s" in language classes factory' %
-                        (assoc, field1, field2)
-                )
-            association = assoc_type_class()
+            association = ModelAssociation(
+                name=f"Assoc_{assoc}",
+                metaconcept=assoc,
```
mrkickling, line 765: And `assoc_type` instead of `metaconcept`

maltoolbox/model.py
```diff
@@ -864,4 +818,33 @@ def load_from_file(
...
+    def __getattr__(self, name):
+        if name in ["info"]:
+            return self.lg_asset.info
+        if (astep := self.lg_asset.attack_steps.get(name, None)):
+            if astep.type == 'defense':
+                return 1. if astep.ttc and astep.ttc['name'] == 'Enabled' else 0.
+        raise AttributeError
+
+class ModelAssociation:
+    def __init__(self, *, name, metaconcept, **kwargs):
```
mrkickling, line 842: Why use _properties here? Wouldn't it be cleaner to have `left_field` and `right_field` that each contains a field with a name and asset type or something?

mrkickling: I also am hesistant to self.__dict__ |= kwargs. What is the reasoning? I like it if we are explicit about all the fields an object can have.

nkakouros: The original PR was still WIP in the sense that, while it worked, things could be rewritten and simplified more. I **think** that:

- I had `_properties` there as a stepping stone to removing the schema objects which had such an attribute and the rest of the code got left_field, etc from there. I.e. yes, I agree `_property` should go away and the rest of the code needs adjusting.

- `self.__dict__ |= kwargs` was a way to have arbitrary kwargs be supported as object attributes directly. E.g. if you pass the arg `my_custom_arg=1` when creating a ModelAsset, then the created object will have `asset.my_custom_arg`.

You had commented in the original PR that this solution still uses an object factory. Do you still think that? I mean, I had in mind to have a factory but I don't remember if I did implemented as such and now I cannot see where the factory logic comes in.

maltoolbox/model.py
andrewbwm: It has already been renamed, sorry, this is my fault, I should have left this as a draft. This is not ready for review.

But on this specific point for future reference. I hate the name metaconcept with a raging passion, it means absolutely nothing, it so devoid of meaning that I cannot fathom while it was ever used. So, if you ever see it in other parts of the code base just rename it to something more relevant.

maltoolbox/model.py
andrewbwm: This pull request should have been a draft, it is not ready for review right now. My mistake.

This will remain lg_asset because I like having a clear description of what the link between instance model and language graph assets/associations is. However, we are very likely to have one of those property methods that just returns the type as a string since a lot of people will use it often.

andrewbwm:
I have a rather drastic proposal that I do actually think makes a lot of sense.

I am not convinced that the `ModelAssociation` class is necessary at all. As you will see from the code included there is a lot of ugly and cumbersome handling of associations trying to figure out which side, left or right, matches a particular asset or field and then also checking to see if they are in the same order as the language graph sometimes. So, there is a good reason to want it gone.

Associations as a concept are purely relational too, they do not mean anything on their own. So, even conceptually it makes sense to remove them.

I think we could simply have the `ModelAsset` have a dictionary called `associations` that is keyed by the fieldname that is applicable to it that just points to the other `ModelAssets` that it is in association with. I think this makes a lot of sense logically and will make the whole thing much more legible both in code, but maybe even in the model format.

maltoolbox/__main__.py
```diff
@@ -51,21 +52,21 @@ def generate_attack_graph(
             log_configs['attackgraph_file']
         )
 
-    if send_to_neo4j:
```
mrkickling, line 54: Why uncomment this?

andrewbwm:
Another consideration is that the AttackerAttachment right now doesn't use the same `_to_dict/_from_dict` pattern which I think it would benefit from.

maltoolbox/__main__.py
andrewbwm: Because neo4j doesn't currently work anymore.

maltoolbox/model.py
```diff
@@ -864,4 +723,193 @@ def load_from_file(
...
+            lg_asset = lg_asset
+        )
+
+        if 'extras' in asset_object:
+            asset.extras = asset_object['extras']
+
+        for defense, defense_value in asset_object.get(
+                'defenses', {}).items():
+            asset.defenses[defense] = float(defense_value)
+
```
nkakouros, line 799:
```suggestion
        asset = ModelAsset(
            name = f"{asset_object['name']}",
            lg_asset = lg_asset,
            defenses = {defense: float(value) for defense, value in asset_object.get('defenses', {}).items()},
            extras = asset_object.get('extras'),
        )
```

Treating defenses and extras differently than properties that you can set via `__init__` args seems arbitrary.

maltoolbox/model.py
```diff
@@ -864,4 +723,193 @@ def load_from_file(
...
+        self.lg_asset:LanguageGraphAsset = lg_asset
+        self.defenses:dict = {}
+        self.extras:dict = {}
+        self.associations:list = []
+        self.attack_step_nodes:list = []
+        for step in self.lg_asset.attack_steps.values():
+            if step.type == 'defense':
+                self.defenses[step.name] = 1.0 if step.ttc and \
+                    step.ttc['name'] == 'Enabled' else 0.0
+
```
nkakouros: So, we are moving away from the one-class-per-asset approach? Are all model assets going to be of type ModelAsset? Is this preferred?

maltoolbox/model.py
```diff
@@ -19,10 +20,8 @@
 
 if TYPE_CHECKING:
     from typing import Any, Optional, TypeAlias
```
mrkickling, line 21: TypeAlias is unused import

maltoolbox/model.py
andrewbwm: Yes, having them be different classes doesn't provide us with anything and it just makes certain checks harder.

nkakouros:
> I have a rather drastic proposal that I do actually think makes a lot of sense.
> 
> I am not convinced that the `ModelAssociation` class is necessary at all. As you will see from the code included there is a lot of ugly and cumbersome handling of associations trying to figure out which side, left or right, matches a particular asset or field and then also checking to see if they are in the same order as the language graph sometimes. So, there is a good reason to want it gone.
> 
> Associations as a concept are purely relational too, they do not mean anything on their own. So, even conceptually it makes sense to remove them.
> 
> I think we could simply have the `ModelAsset` have a dictionary called `associations` that is keyed by the fieldname that is applicable to it that just points to the other `ModelAssets` that it is in association with. I think this makes a lot of sense logically and will make the whole thing much more legible both in code, but maybe even in the model format.

This makes sense. The only thing I can think of is if the mal-gui or the simulator do have a need for separate association objects and removing them would make them slower or their code uglier by having to loop assets, their associations, etc.

maltoolbox/model.py
nkakouros: #99

nkakouros: :-D

maltoolbox/model.py
```diff
@@ -864,4 +723,193 @@ def load_from_file(
...
         else:
             raise ValueError('Unknown file extension, expected json/yml/yaml')
-        return cls._from_dict(serialized_model, lang_classes_factory)
+        return cls._from_dict(serialized_model, lang_graph)
+
+
+class ModelAsset:
+    def __init__(self, name:str, lg_asset: LanguageGraphAsset):
+        self.name:str = name
+        self.id:Optional[int] = None
```
mrkickling, line 731: nitpick, but space after the ':' helps for readability

maltoolbox/model.py
```diff
@@ -864,4 +723,193 @@ def load_from_file(
...
+            self.lg_association.right_field.fieldname: right_assets
+        })
+
+
+    @cached_property
+    def type(self):
+        return self.lg_association.name
+
+
+    def get_association_field_names(
```
mrkickling, line 822: Can be called just `get_fieldnames`

maltoolbox/model.py
```diff
@@ -864,4 +722,186 @@ def load_from_file(
...
+        self.defenses:dict = {}
+        self.extras:dict = {}
+        self.associations:list = []
+        self.attack_step_nodes:list = []
+        for step in self.lg_asset.attack_steps.values():
+            if step.type == 'defense':
+                self.defenses[step.name] = 1.0 if step.ttc and \
+                    step.ttc['name'] == 'Enabled' else 0.0
+
+
```
nkakouros, line 743: Making this a dataclass will simplify the code and give access to ready implementations of `__repr__` and `__str__`, and also support `asdict` that could simplify `_to_dict` in the future.

maltoolbox/model.py
```diff
@@ -864,4 +722,186 @@ def load_from_file(
             serialized_model = load_dict_from_json_file(filename)
         else:
             raise ValueError('Unknown file extension, expected json/yml/yaml')
-        return cls._from_dict(serialized_model, lang_classes_factory)
+        return cls._from_dict(serialized_model, lang_graph)
+
+
+class ModelAsset:
+    def __init__(self, name:str, lg_asset: LanguageGraphAsset):
```
nkakouros, line 729: Do we want to make by default all our classes hashable so that we don't have to deal with that if ever needed? It will also make explicit what we consider the id of every object type to be.

maltoolbox/model.py
andrewbwm: The reasoning is that they're not mandatory, but I do see your point. I also considered adding the defenses as a parameter for the construct because I do think there are many situations where it's most convenient to set the defenses then. So, I'll add both for completion, but the extras are highly optional.

maltoolbox/model.py
andrewbwm: Not happening, I've had my full of dataclasses. :)

maltoolbox/model.py
nkakouros: You can set the optional stuff in `__init__` as args with default values.

maltoolbox/model.py
```diff
@@ -864,4 +723,193 @@ def load_from_file(
             serialized_model = load_dict_from_json_file(filename)
         else:
             raise ValueError('Unknown file extension, expected json/yml/yaml')
-        return cls._from_dict(serialized_model, lang_classes_factory)
+        return cls._from_dict(serialized_model, lang_graph)
+
+
+class ModelAsset:
```
mrkickling, line 728: Are we sure the ModelAsset needs an ID?
I see it is used as a key in the dict in model, I was thinking if the name could have been used as key instead, but maybe it is not necessarily unique?

If we use ID, then it might be a good idea to have it not be optional and part of arguments to `__init__`? Or would it make sense to ever have an asset without an ID?

maltoolbox/model.py
andrewbwm: Do you mean implement a `__hash__` method that returns the id? If so, we can do that, but they might not have an id which complicates things potentially.

maltoolbox/model.py
```diff
@@ -19,10 +20,8 @@
 
 if TYPE_CHECKING:
     from typing import Any, Optional, TypeAlias
-    from .language import LanguageClassesFactory
-    from python_jsonschema_objects.classbuilder import ProtocolBase
-
-    SchemaGeneratedClass: TypeAlias = ProtocolBase
+    from .language import (LanguageGraph, LanguageGraphAsset,
```
mrkickling, line 22: I think we should have the imports on a new line when using parenthesis.
It might be something to decide later for the whole project so you can ignore this comment if you don't agree

maltoolbox/model.py
nkakouros: Let's leave this for when hashable model assets are actually needed.

maltoolbox/model.py
```diff
@@ -864,4 +722,186 @@ def load_from_file(
...
+
+class ModelAssociation:
+    def __init__(self,
+            lg_assoc: LanguageGraphAssociation,
+            left_assets: list = [],
+            right_assets: list = []):
+        self.lg_association:LanguageGraphAssociation = lg_assoc
+        self.type = self.lg_association.name
+        self.extras:dict = {}
+        self.__dict__.update({
```
mrkickling, line 815: hm, what is the use of `__dict__`?  I have not seen it used before.

andrewbwm: These are the object properties, so the fields are added a object properties, I think it's equivalent to `setattr`.

maltoolbox/model.py
nkakouros: We can also add a `right_field`, `left_field` property to the object directly. Or a `field_names` property.

But the association class is going away right?

maltoolbox/model.py
mrkickling: Got it. 
So we still would have to do `get_field_names` and then get the specific field name one at a time. I don't really like that (this is one thing i have disliked with the langclasses factory)

Another idea would be to have `self.left_field` containing a `field_name` and an `asset` and a `self.right_field` containing a `field_name` and an `asset` (perhaps type also). Then you can just write `self.left_field.asset`, `self.right_field.asset` to fetch associated assets instead of having to get the field names first. 

Another option is to remove the Associations altogether from the model as you mentioned, I totally agree with that. But we have to decide whether it should be done now or in the future. If we postpone it we should make sure we are happy with the ModelAssociation that we merge from this PR first.

nkakouros:
I just tried this branch with the mal simulator and it worked great. If I am not delusional this greatly improved speed as well! The rewrite from lists to sets in the expression chains brought it down to ~1 minute instead of 2 hours and this here brings it to 2 seconds from 1 minute! 👏

maltoolbox/model.py
mrkickling: If they are going away in this PR all my comments on the ModelAssoc can be ignored:P

maltoolbox/__main__.py
mrkickling: Got it. But I think you should instead raise an exception 'NEO4j does not work' so users don't go mental when their graphs don't show up.

andrewbwm: I will one of us will fix it at some later point and I don't think enough people use Neo4j and keep their mal-toolbox up to date for it to matter.

maltoolbox/model.py
```diff
@@ -728,11 +728,18 @@ def load_from_file(
...
-        defenses: dict = {},
-        extras: dict = {}):
+        defenses: Optional[dict] = None,
+        extras: Optional[dict] = None
+    ):
+
+        if defenses is None:
+            defenses = {}
+        if extras is None:
+            extras = {}
```
nkakouros, line 742:
```suggestion
        self.defenses = defenses or {}
        self.extras = extras or {}
```

maltoolbox/model.py
```diff
@@ -864,4 +477,174 @@ def load_from_file(
...
+    def add_associated_assets(self, fieldname: str, assets: list[ModelAsset]):
+        """ Add the assets provided as a parameter to the set of associated
+        assets dictionary entry corresponding to the fieldname parameter.
+        """
+
+        # Add the associated assets to this assets dictionary
+        self.validate_new_association(fieldname, assets)
+        self.associated_assets.setdefault(
+            fieldname, set()
+        ).update(set(assets))
```
nkakouros, line 622:
```suggestion
        ).update(assets)
```

maltoolbox/model.py
```diff
@@ -864,4 +477,174 @@ def load_from_file(
...
+
+        # Also add this asset to the associated assets' dictionaries
+        lg_assoc = self.lg_asset.associations[fieldname]
+        other_fieldname = lg_assoc.get_opposite_fieldname(fieldname)
+
+        for asset in assets:
+            asset.validate_new_association(other_fieldname, [self])
+            asset.associated_assets.setdefault(
+                other_fieldname, set()
+            ).add(self)
```
nkakouros, line 632: This will fail fine if any of these validations fails. But the user may catch that expression and say "hey its fine if I tried connecting that asset to this one and I couldn't because there were too many already" (e.g. mal-gui may do that). In this case, if the line that fails is 629, then line 620 will have already added the one direction of the association, leaving the model in an inconsistent state. What I am saying is that it's better if this "add_associated_assets" method became atomic and the user gets either both directions of the association added or none. For which i think it would be simpler to have this be singular (`add_assocated_asset`).

maltoolbox/model.py
```diff
@@ -864,4 +477,174 @@ def load_from_file(
...
+        if fieldname not in self.lg_asset.associations:
+            accepted_fieldnames = list(self.lg_asset.associations.keys())
+            raise LookupError(
+                f"Fieldname '{fieldname}' is not an accepted association "
+                f"fieldname from asset type {self.lg_asset.name}. "
+                f"Did you mean one of {accepted_fieldnames}?"
+            )
+
+        # Get the field that the fieldname belongs to
+        # TODO: lg_assoc.get_field(fieldname) would be nice to have for this
```
nkakouros, line 596: Why not add it?

maltoolbox/model.py
```diff
@@ -864,4 +477,174 @@ def load_from_file(
...
+        is valid with the assets given in param `assets_to_add`:
+        - fieldname is valid for the asset type of this ModelAsset
+        - Make sure no more assets than 'field.maximum' are added to the field
+
+        Raises:
+            LookupError - fieldname can not be found for this ModelAsset
+            ValueError - there will be too many assets in the field
+                         if we add this association
+
+        TODO: is there anything else we want to validate?
```
nkakouros, line 583: Do we care about the asset type being added? The association may expect a Computer for the 'hosts' field and the user may pass a Lawnmower.

maltoolbox/model.py
```diff
@@ -864,4 +477,174 @@ def load_from_file(
...
+                asset_object.get('defenses', {}).items()},
+            extras = asset_object.get('extras', {})
+        )
+
+        return asset
+
+    def __repr__(self):
+        return self.name
+
+    def validate_new_association(
```
nkakouros, line 569: Do you expect this to be part of the public interface of the class? Makes more sense to be a private method.

maltoolbox/model.py
```diff
@@ -864,4 +477,174 @@ def load_from_file(
...
+        assets_in_field_after = assets_in_field_before | set(assets_to_add)
+        max_assets_in_field = assoc_field.maximum or math.inf
+
+        if len(assets_in_field_after) > max_assets_in_field:
+            raise ValueError(
+                f"You can have maximum {assoc_field.maximum} "
+                f"assets for association field {fieldname}"
+            )
+
+    def add_associated_assets(self, fieldname: str, assets: list[ModelAsset]):
```
nkakouros, line 613: Given that the expected way to add associated assets is this method, it makes sense to mark the underlying `associated_assets` property as read only making it private and adding an `@property` public version.

maltoolbox/model.py
nkakouros: Same for remove_associations.

maltoolbox/model.py
```diff
@@ -232,7 +232,7 @@ def remove_asset(self, asset: ModelAsset) -> None:
 
         # First remove all of the associated assets
         for fieldname, assoc_assets in asset.associated_assets.items():
-            asset.remove_associated_assets(fieldname, assoc_assets)
+            asset.remove_associated_assets(fieldname, list(assoc_assets))
```
mrkickling, line 235: What is the reason for 'list(assoc_assets)' ?

maltoolbox/model.py
mrkickling: Agreed

maltoolbox/model.py
mrkickling: Very good point, nice catch

maltoolbox/model.py
mrkickling: We talked about this, and I think we do. But it will already fail if the fieldname is not supported in the language for the asset type so currently we catch most of the errors. But we should definitely add a more explicit type check.

maltoolbox/model.py
mrkickling: Yes, but since it was called on something else than self in `add_associated_assets` I had it be public, but I think the solution is to rethink that method. Definitely want it private.

maltoolbox/model.py
mrkickling: Makes sense

maltoolbox/model.py
mrkickling: Fixed

maltoolbox/model.py
mrkickling: Fixed

maltoolbox/model.py
mrkickling: Fixed

maltoolbox/model.py
mrkickling: Added type checking

maltoolbox/language/languagegraph.py
```diff
@@ -616,6 +625,11 @@ def __init__(self, lang: Optional[dict] = None):
             self._generate_graph()
 
 
+    @cached_property
+    def associations(self):
+        return {k: v for asset in self.assets.values() for k, v in asset.associations.items()}
+
+
```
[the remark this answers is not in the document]
mrkickling: who added this? :angry:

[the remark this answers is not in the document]
andrewbwm: A very foolish old man.

maltoolbox/model.py
mrkickling: Changed so that validation is done for both sides before adding any side of the association. Might be nicer ways to do this like atomic solution you proposed, but we will not care about that right now.

maltoolbox/model.py
mrkickling: Private now

maltoolbox/model.py
mrkickling: fixed, and also id is property now

mrkickling (approved):

maltoolbox/model.py
andrewbwm: @mrkickling and I decided to not have it be a dataclass just yet because of the default defense values handling. We might change it to a dataclass down the line if we feel it's a particularly good fit.

maltoolbox/model.py
```diff
@@ -129,38 +131,43 @@ def remove_entry_point(
...
+            lang_graph: LanguageGraph,
             mt_version: str = __version__
         ):
 
         self.name = name
-        self.assets: list[SchemaGeneratedClass] = []
-        self.associations: list[SchemaGeneratedClass] = []
+        self.assets: dict[int, ModelAsset] = {}
+        self._name_to_asset:dict[str, ModelAsset] = {} # optimization
         self._type_to_association:dict = {} # optimization
```
mrkickling, line 151: Can be removed

PR #104, opened 2025-01-24
Support loading older version model file

mrkickling:
~Support loading older version model file.~

~When running wrappers.create_attack_graph, it will try to load the model normally.
If this does not work, it will try to load it from an old format.
If that does not work, it will crash.~

EDIT:
- Support converting/upgrading old models with maltoolbox cli
- Support converting models from 0.1.0 -> 0.2.0
- Give suggestion to convert if model can not load.


nkakouros:
I think we should support only the latest format of everything. For older formats, we could provide a simple standalone script like:

```
def convert_from_securicad_to_maltoolbox_v1(): ...

def convert_from_maltoolbox_v1_to_maltoolbox_v2(): ...

def convert_from_maltoolbox_v2_to_maltoolbox_v3(): ...
```

This way the main code stays clean and the conversion logic is clean. If someone has a very old file, the script will just run the chain of methods needed.

mrkickling:
> I think we should support only the latest format of everything. For older formats, we could provide a simple standalone script like:
> 
> ```
> def convert_from_securicad_to_maltoolbox_v1(): ...
> 
> def convert_from_maltoolbox_v1_to_maltoolbox_v2(): ...
> 
> def convert_from_maltoolbox_v2_to_maltoolbox_v3(): ...
> ```
> 
> This way the main code stays clean and the conversion logic is clean. If someone has a very old file, the script will just run the chain of methods needed.

You mean we do not want it to happen automatically but instead provide a script for it?

mrkickling:
> I would say so. The main code becomes bloated otherwise. And this process should only happen once. If anyone has old file formats, they can run the converter and update them, why should they keep the old formats around?

Makes sense. I think this should be part of the maltoolbox cli in that case. What do you think?

mrkickling:
> > I think we should support only the latest format of everything. For older formats, we could provide a simple standalone script like:
> > ```
> > def convert_from_securicad_to_maltoolbox_v1(): ...
> > 
> > def convert_from_maltoolbox_v1_to_maltoolbox_v2(): ...
> > 
> > def convert_from_maltoolbox_v2_to_maltoolbox_v3(): ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This way the main code stays clean and the conversion logic is clean. If someone has a very old file, the script will just run the chain of methods needed.
> 
> I also suggested chaining them. While it's a bit more cumbersome for really old formats, we shouldn't have that many of them laying around and otherwise we would have to update the updaters every time we change the format.
> 
> I would also suggest that we don't need to go too far back or fill in all of the gaps because there were too many intermediary formats we've had.
> 
> The last thing to mention is that it is probably ok to not support some really old formats and just spit out an error message that it cannot be handled.
> 
> Whether they're separate scripts or not isn't really of great importance to me.

Yes, to me the important part is to make it easy to convert models to the new format. If it is automatic or not doesn't matter (but they should easily see why the toolbox is crashing if it is because of an old model).

nkakouros:
How about this goes to model.convert? and gets exposed as a cli command?

mrkickling:
> How about this goes to model.convert? and gets exposed as a cli command?

You mean as a static method in maltoolbox.Model? Like .load_from_old_format() ?
I think it works better like a standalone module for updating models, but I might not understand how you mean it should work.

I now changed it so that model upgrading is part of the maltoolbox CLI instead of happening automatically.
If the model fails to parse it will add an exception that gives the user a hint to try `maltoolbox upgrade-model` if they have an old model.

I also added `maltoolbox` as a script in pyproject.toml (I had to run it as `python3 -m maltoolbox` otherwise, did you as well?

I think this feature can be worked on more, but I also think it would be nice for the users to have this tool at least before the new mal-simulator uses the new mal-toolbox.

nkakouros:
Sounds good. I would only do this differently. I would have the converted be a module like:

```
model/
  __init__.py
  convert.py
```

where `__init__` holds the current model.py contents. I would not expose anything from convert.py higher, i.e. you will always need to do `from model.convert import something`, `from model import something_from_convert` would not be possible. This would allow us to delete the translators folder and keep the structure simpler.

mrkickling:
> Sounds good. I would only do this differently. I would have the converted be a module like:
> 
> ```
> model/
>   __init__.py
>   convert.py
> ```
> 
> where `__init__` holds the current model.py contents. I would not expose anything from convert.py higher, i.e. you will always need to do `from model.convert import something`, `from model import something_from_convert` would not be possible. This would allow us to delete the translators folder and keep the structure simpler.

Then I understand, having converter/updater as a submodule of model makes sense to me.
However, I feel against having classes defined in __init__, it hides things. Only things in there should be exports of the package.

nkakouros:
It's a matter of interpretation. To me the `__init__` file is a way to add content on a folder. When you do `import module` and the module is defined in a module.py file, then that file is loaded. What happens when you do `import module` and `module` is a folder? Folders do not hold text themselves. Here comes `__init__.py`, holding the text. It's also redundant to do:

```
from model.model import Model
```
and have a model/model.py file. Or to have such a file and import `Model` in `__init__` so that you can then do:

```
from model import Model
```

when you could just have the code in `__init__.py` from the start and save an extra file.

mrkickling:
> It's a matter of interpretation. To me the `__init__` file is a way to add content on a folder. When you do `import module` and the module is defined in a module.py file, then that file is loaded. What happens when you do `import module` and `module` is a folder? Folders do not hold text themselves. Here comes `__init__.py`, holding the text. It's also redundant to do:
> 
> ```
> from model.model import Model
> ```
> 
> and have a model/model.py file. Or to have such a file and import `Model` in `__init__` so that you can then do:
> 
> ```
> from model import Model
> ```
> 
> when you could just have the code in `__init__.py` from the start and save an extra file.

I would chose the approach to have a `./model/model.py`, a `./model/updater.py` and a `./model/__init__.py` and the `__iniy__.py` would import Model from model and whatever we want to be public in the updater. And it would contain the `__all__`-list that shows which modules are imported with import *.  So then you could do `from model import Model` without having to define any class in `__init__.py`.

I think having the extra file is a good thing and not a drawback with this solution. Otherwise the next person looking into the package has to realize that classes are defined in the __init__, which is not an all too common practice. Especially if the package grows, having separate files helps I think.

nkakouros:
If I am not mistake a lot if not most of the packages in the python stdlib are written in __init__.py directly. The recommendation also seems to be to keep file structure as flat as possible, as compared to eg java where it is file madness. But whatever you prefer as the maintainers.

mrkickling:
You are right, I looked it up now. Interesting, what I have learned it is a big 'no, no'..
I looked at numpy and fastapi and they define meta things and imports in `__init__.py` which is what I am also used to.

I still think it is better to not define things in `__init__.py` for readability/maintainability. I guess I could feel differently if it is a very small package (but then I feel it should just be a .py-file/module instead of a directory). With that I am not saying we need java style with one file per class.

nkakouros:
Checkout this session 

https://www.youtube.com/watch?v=wf-BqAjZb8M

Generally, all of his talks are great.

mrkickling:
Could this branch me merged as it is though and we could wait until deciding whether to make model a package with updater/converter as submodule? @nkakouros

nkakouros:
LGTM

nkakouros:
@joakim Is the use case that the user will run the convert command and then no matter what the old version is the model file will be rewritten to latest version? 

A suggestion is to inform in the toolbox if an old model file is to be loaded and fail with an informative error. Although, now that the conversion script is in the maltoolbox package, we can call the functions for automatic conversion...

mrkickling:
Well, my initial solution was to do it automatically. But I changed it to be an informative error message after your previous feedback. 

Currently this is the approach:
- User loads old model
- Exception is raised and user is told to try `maltoolbox upgrade-model` 
- User runs `maltoolbox upgrade-model <model_file> <lang_file> -o <output-file>`
- User loads new model

It will not necessarily override, the user can decide what the output file name is.

A more automatic conversion approach might be nice like:
- User loads old model
- User is told the model might be old. 'Do you want to override the model with an upgraded version?'
- User says yes or no and the model is overridden or not.
- User loads new model

mrkickling:
> Alternatively, the code could just convert in memory and never save.

Yes, but then we will have some users that never update there models so we would have to support them forever (not a big problem if the converter is chained as I now implemented it as I guess). But I think there should at least be a warning or question about converting it to latest version.

Btw, i rebased on top of `drop-jsonschema` and implemented a converter for the model format on that branch.

mrkickling:
I think for now we can be conservative and be happy with an error message and suggestion to convert, it will be enough for most cases imo.

This is what it will look like when you try to load an older model:
```
maltoolbox.exceptions.ModelException: Could not load model. It might be of an older version. Try to upgrade it with 'maltoolbox upgrade-model'
````

andrewbwm (approved):

PR #105, opened 2025-01-24
 Add LookupDict for faster and simpler lookups

andrewbwm:
This is a rebased version of #95. See comments there for the state at this particular point in time.

### DIFF
--- a/maltoolbox/attackgraph/attackgraph.py
+++ b/maltoolbox/attackgraph/attackgraph.py
@@ -32,7 +32,7 @@ logger = logging.getLogger(__name__)
 class AttackGraph():
     """Graph representation of attack steps"""
     def __init__(self, lang_graph, model: Optional[Model] = None):
-        self.nodes: LookupDict[str, AttackGraphNode] = LookupDict()
+        self.nodes: LookupDict[int, AttackGraphNode] = LookupDict()
         self.attackers: list[Attacker] = []
         # Dictionaries used in optimization to get nodes and attackers by id
         # or full name faster
@@ -660,7 +660,7 @@ class AttackGraph():
             child.parents.remove(node)
         for parent in node.parents:
             parent.children.remove(node)
-        self.nodes.pop(node.id)  # type: ignore[call-overload]
+        self.nodes.pop(node.id)  # type: ignore[call-overload,arg-type]
 
         if not isinstance(node.id, int):
             raise ValueError(f'Invalid node id.')
--- a/maltoolbox/attackgraph/attackgraph.py
+++ b/maltoolbox/attackgraph/attackgraph.py
@@ -241,21 +241,6 @@ class AttackGraph():
         return cls._from_dict(serialized_attack_graph,
             lang_graph, model = model)
 
-    def get_node_by_full_name(self, full_name: str) -> Optional[AttackGraphNode]:
-        """
-        Return the attack node that matches the full name provided.
-
-        Arguments:
-        full_name   - the full name of the attack graph node we are looking
-                      for
-
-        Return:
-        The attack step node that matches the given full name.
-        """
-
-        logger.debug(f'Looking up node with full name "%s"', full_name)
-        return self._full_name_to_node.get(full_name)
-
     def get_attacker_by_id(self, attacker_id: int) -> Optional[Attacker]:
         """
         Return the attacker that matches the id provided.
@@ -302,7 +287,7 @@ class AttackGraph():
             for (asset, attack_steps) in attacker_info.entry_points:
                 for attack_step in attack_steps:
                     full_name = asset.name + ':' + attack_step
-                    ag_node = self.get_node_by_full_name(full_name)
+                    ag_node = self.nodes.fetch("full_name", full_name)
                     if not ag_node:
                         logger.warning(
                             'Failed to find attacker entry point '
@@ -593,8 +578,8 @@ class AttackGraph():
                             if target_asset is not None:
                                 target_node_full_name = target_asset.name + \
                                     ':' + target_attack_step.name
-                                target_node = self.get_node_by_full_name(
-                                    target_node_full_name)
+                                target_node = self.nodes.fetch("full_name",
+                                     target_node_full_name)
                                 if target_node is None:
                                     msg = ('Failed to find target node '
                                            '"%s" to link with for attack '
--- a/maltoolbox/attackgraph/attackgraph.py
+++ b/maltoolbox/attackgraph/attackgraph.py
@@ -173,8 +173,8 @@ class AttackGraph():
             attack_graph.add_node(ag_node, node_id=node_dict['id'])
 
         # Re-establish links between nodes.
-        for node_dict in serialized_attack_steps.values():
-            _ag_node = attack_graph.get_node_by_id(node_dict['id'])
+        for node_full_name, node_dict in serialized_attack_steps.items():
+            _ag_node = attack_graph.nodes[node_dict['id']]
             if not isinstance(_ag_node, AttackGraphNode):
                 msg = ('Failed to find node with id %s when loading'
                        ' attack graph from dict')
@@ -182,7 +182,7 @@ class AttackGraph():
                 raise LookupError(msg % node_dict["id"])
             else:
                 for child_id in node_dict['children']:
-                    child = attack_graph.get_node_by_id(int(child_id))
+                    child = attack_graph.nodes[int(child_id)]
                     if child is None:
                         msg = ('Failed to find child node with id %s'
                                ' when loading from attack graph from dict')
@@ -191,7 +191,7 @@ class AttackGraph():
                     _ag_node.children.append(child)
 
                 for parent_id in node_dict['parents']:
-                    parent = attack_graph.get_node_by_id(int(parent_id))
+                    parent = attack_graph.nodes[int(parent_id)]
                     if parent is None:
                         msg = ('Failed to find parent node with id %s '
                                'when loading from attack graph from dict')
@@ -240,20 +240,6 @@ class AttackGraph():
             raise ValueError('Unknown file extension, expected json/yml/yaml')
         return cls._from_dict(serialized_attack_graph,
             lang_graph, model = model)
-
-    def get_node_by_id(self, node_id: int) -> Optional[AttackGraphNode]:
-        """
-        Return the attack node that matches the id provided.
-
-        Arguments:
-        node_id     - the id of the attack graph node we are looking for
-
-        Return:
-        The attack step node that matches the given id.
-        """
-
-        logger.debug('Looking up node with id %s', node_id)
-        return self._id_to_node.get(node_id)
 
     def get_node_by_full_name(self, full_name: str) -> Optional[AttackGraphNode]:
         """
@@ -739,7 +725,7 @@ class AttackGraph():
 
         self.next_attacker_id = max(attacker.id + 1, self.next_attacker_id)
         for node_id in reached_attack_steps:
-            node = self.get_node_by_id(node_id)
+            node = self.nodes[node_id]
             if node:
                 attacker.compromise(node)
             else:
@@ -748,7 +734,7 @@ class AttackGraph():
                 logger.error(msg, node_id)
                 raise AttackGraphException(msg % node_id)
         for node_id in entry_points:
-            node = self.get_node_by_id(int(node_id))
+            node = self.nodes[int(node_id)]
             if node:
                 attacker.entry_points.append(node)
             else:
--- a/maltoolbox/attackgraph/attackgraph.py
+++ b/maltoolbox/attackgraph/attackgraph.py
@@ -2,11 +2,10 @@
 MAL-Toolbox Attack Graph Module
 """
 from __future__ import annotations
-import copy
 import logging
 import json
 
-from itertools import chain
+from copy import deepcopy
 from typing import TYPE_CHECKING
 
 from .node import AttackGraphNode
@@ -21,6 +20,7 @@ from ..file_utils import (
     load_dict_from_yaml_file,
     save_dict_to_file
 )
+from ..utils import LookupDict
 
 
 if TYPE_CHECKING:
@@ -32,7 +32,7 @@ logger = logging.getLogger(__name__)
 class AttackGraph():
     """Graph representation of attack steps"""
     def __init__(self, lang_graph, model: Optional[Model] = None):
-        self.nodes: list[AttackGraphNode] = []
+        self.nodes: LookupDict[str, AttackGraphNode] = LookupDict()
         self.attackers: list[Attacker] = []
         # Dictionaries used in optimization to get nodes and attackers by id
         # or full name faster
@@ -54,7 +54,7 @@ class AttackGraph():
         """Convert AttackGraph to dict"""
         serialized_attack_steps = {}
         serialized_attackers = {}
-        for ag_node in self.nodes:
+        for ag_node in self.nodes.values():
             serialized_attack_steps[ag_node.full_name] =\
                 ag_node.to_dict()
         for attacker in self.attackers:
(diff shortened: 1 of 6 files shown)


nkakouros:
I also think that it makes sense to merge file_utils.py into utils.py.

mrkickling (approved):

maltoolbox/attackgraph/attackgraph.py
```diff
@@ -82,15 +83,13 @@ def create_attack_graph(
...
+        self.nodes: LookupDict[int, AttackGraphNode] = LookupDict()
         self.attackers: list[Attacker] = []
         # Dictionaries used in optimization to get nodes and attackers by id
         # or full name faster
-        self._id_to_node: dict[int, AttackGraphNode] = {}
-        self._full_name_to_node: dict[str, AttackGraphNode] = {}
         self._id_to_attacker: dict[int, Attacker] = {}
 
-        self.model = model
+        self.model =model
```
mrkickling, line 92: missing whitespace

maltoolbox/attackgraph/attackgraph.py
```diff
@@ -124,40 +123,24 @@ def __deepcopy__(self, memo):
...
-        for node in self.nodes:
+        for node in self.nodes.values():
+            nid = id(node)
             if node.parents:
-                memo[id(node)].parents = copy.deepcopy(node.parents, memo)
+                memo[nid].parents = deepcopy(node.parents, memo)
             if node.children:
-                memo[id(node)].children = copy.deepcopy(node.children, memo)
+                memo[nid].children = deepcopy(node.children, memo)
+            if node.compromised_by:
```
mrkickling, line 139: Do we have tests for this?

mrkickling: The reason why compromised_by was done after the nodes had been created was so that there would not be too deep recursion I believe, so run deepcopy on a very large graph to make sure.

maltoolbox/attackgraph/attackgraph.py
nkakouros:
```suggestion
        self.model = model
```

#99 !!

maltoolbox/attackgraph/attackgraph.py
nkakouros: > Do we have tests for this?

I haven't added any, I don't know.



> The reason why compromised_by was done after the nodes had been created was so that there would not be too deep recursion I believe, so run deepcopy on a very large graph to make sure.

You may be into sth here, but I fail to see it in the code. The way I see the changed code is that it just copies objects around. `memo` should take care of performance and guarantee that each object is copied only once.

maltoolbox/utils.py
```diff
@@ -0,0 +1,59 @@
...
+    def fetch(self, key: str, value: Any, op: str="eq") -> Any:
+        ret = self.lookup(key, value, op)
+
+        if len(ret) > 1:
+            raise ValueError(f"Multiple items match {key} ~ {op} ~ {value}")
+
+        try:
+            return next(iter(ret))
+        except KeyError:
+            return None
```
nkakouros, line 59: Regarding documentation this is what chatgpt suggested, with minor fixes from me. I also removed the dataclass import that was unused. Press accept suggestion if you agree.

```suggestion
"""Various generic classes and methods."""
import operator
from typing import Any, Dict, TypeVar

K = TypeVar("K")
V = TypeVar("V")

class LookupDict(Dict[K, V]):
    """
    A dict subclass that provides lookup functionality based on object attributes.
    
    Maintains an internal index to optimize repeated lookups based on a specific key.
    """
    
    def __init__(self, *args, **kwargs):
        """
        Initialize the LookupDict with optional initial dictionary data.
        
        Arguments should be passes similar to `dict()`.
        
        Args:
            *args: Positional arguments passed to the base dictionary.
            **kwargs: Keyword arguments passed to the base dictionary.
        """
        super().__init__(*args, **kwargs)
        self._indices = {}

    def __setitem__(self, key, value):
        """
        Set an item in the dictionary and clear all cached indices.
        
        Args:
            key (K): The key under which the value is stored.
            value (V): The value to store.
        """
        super().__setitem__(key, value)
        self._indices.clear()

    def __delitem__(self, key):
        """
        Delete an item from the dictionary and clear all cached indices.
        
        Args:
            key (K): The key to delete.
        """
        super().__delitem__(key)
        self._indices.clear()

    def lookup(self, key: str, value: Any, op: str ="eq") -> set[Any]:
        """
        Perform a lookup in the dictionary based on an attribute and a comparison operation.
        
        Args:
            key (str): The attribute name to look up within the stored values.
            value (Any): The value to compare against.
            op (str, optional): The comparison operator (default: "eq").
                Supported operators include "eq", "in", and others from the `operator` module.
        
        Returns:
            set[Any]: A set of matching values from the dictionary.
        
        Raises:
            ValueError: If an invalid operator is provided.
        """
        if key not in self._indices:
            self._indices[key] = {getattr(v, key): set() for v in self.values()}
            for v in self.values():
                if not (vkey := getattr(v, key)) in self._indices[key].keys():
                    self._indices[key][vkey] = set()
                self._indices[key][vkey].add(v)

        if op == "eq":
            try:
                return self._indices[key][value]
            except KeyError:
                return set()

        if op == "in":
            oper = lambda a, b: operator.contains(b, a)
        else:
            try:
                oper = getattr(operator, op)
            except AttributeError:
                raise ValueError(f"Invalid operator {op}")

        ret = set()
        for i in filter(lambda k: oper(k, value), self._indices[key]):
            ret.update(self._indices[key][i])

        return ret

    def fetch(self, key: str, value: Any, op: str ="eq") -> Any:
        """
        Fetch a single matching item from the dictionary.
        
        If multiple items match, an exception is raised. If you expect
        multiple matches use `lookup()` instead.
        
        Args:
            key (str): The attribute name to look up within the stored values.
            value (Any): The value to compare against.
            op (str, optional): The comparison operator (default: "eq").
                Supported operators include "eq", "in", and others from the `operator` module.
        
        Returns:
            Any: The matching item if one is found, otherwise `None`.
        
        Raises:
            ValueError: If multiple matching items are found.
        """
        ret = self.lookup(key, value, op)

        if len(ret) > 1:
            raise ValueError(f"Multiple items match {key} ~ {op} ~ {value}")

        try:
            return next(iter(ret))
        except KeyError:
            return None
```

nkakouros: @mrkickling

maltoolbox/attackgraph/attackgraph.py
mrkickling: Yeah, personally I find it quite unintuitive so test would be required (maybe it already exists). But I would suggest trying to run deepcopy on a very large attack graph as well.

maltoolbox/attackgraph/attackgraph.py
```diff
@@ -305,35 +289,6 @@ def load_from_file(
...
-        node_id     - the id of the attack graph node we are looking for
-
-        Return:
-        The attack step node that matches the given id.
-        """
-
-        logger.debug('Looking up node with id %s', node_id)
-        return self._id_to_node.get(node_id)
-
-    def get_node_by_full_name(self, full_name: str) -> Optional[AttackGraphNode]:
```
mrkickling, line 322: I personally liked the methods get_node_by_id and get_node_by_full_name. They made it clear what attributes you could fetch a node by. Now the user would do `attack_graph.nodes.fetch(attr, value)` without knowing which attributes are supported. Or are all attributes of nodes supported? If so, could we still have the get_node_by_x and use the LookupDict in those? Or at least have  `AttackGraph.get_node_by_attribute_value(attr: str, value: Any)` so the user does not have to go through `attack_graph.nodes.fetch()` ?

maltoolbox/utils.py
```diff
@@ -0,0 +1,59 @@
...
+from dataclasses import dataclass
+from typing import Any, Dict, TypeVar
+
+K = TypeVar("K")
+V = TypeVar("V")
+
+class LookupDict(Dict[K, V]):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._indices = {}
```
mrkickling, line 11: _indices should probably be called _keys?

maltoolbox/utils.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def __setitem__(self, key, value):
+        super().__setitem__(key, value)
+        self._indices.clear()
```
mrkickling, line 15: Every time you add a new item the keys are cleared so the dicts have to be recalculated on the next lookup?

maltoolbox/utils.py
```diff
@@ -0,0 +1,59 @@
...
+            self._indices[key] = {getattr(v, key): set() for v in self.values()}
+            for v in self.values():
+                if not (vkey:=getattr(v, key)) in self._indices[key].keys():
+                    self._indices[key][vkey] = set()
+
+                self._indices[key][vkey].add(v)
+
+        if op == "eq":
+            try:
+                return self._indices[key][value]
```
mrkickling, line 32: Maybe use .get(value, set()) here?

maltoolbox/utils.py
nkakouros: Each index is a full dictionary stored under a key in the _indices.

mrkickling: Ok, so a key name maps to a dict that maps from keys of that name to values. I still think _indices might not be the best description of it, but it is hard to make up a good name perhaps. A typehint would help at least!

nkakouros: Yes, I will add a type hint. I see where you are coming from, you read index as in a list index, while I had in mind database indices.

maltoolbox/utils.py
nkakouros: Right. I should probably just remove the item from the indices and not flush it all... I will. update. For things like this that may slip we should have performance tracking

maltoolbox/utils.py
nkakouros: Oops, right!

maltoolbox/attackgraph/attackgraph.py
nkakouros: All valid attributes are supported. Removing those methods is what this PR does. Having them back and then going through the lookup dict undoes this. 

I don't like the get methods as:

- they are too java-ish
- they take up lines of code for no reason
- whenever you need to get an item by another attribute that is not yet implemented you have to add new methods. With the lookup dict you can use whatever attribute you like whenever you like. 
- they don't support comparisons other than equals
- each one needs its own unit test
- they make a bug more probable
- they make refactorings and optimizations harder as changes need to be made all over the code
- they don't support plural as in fetch all that match
- they require to bloat the class with as many private attributes.

As long as the user understands what fetch() or lookup() do, they won't have difficulty in reading the code. They are also cognitively more light as you don't have to read and read long lines, long method names etc. You read `nodes.fetch(key, value)` and you understand the same thing as "get node by key"  but via a single consistent API that works across any object and that you read and learn once as compared to reading 10 different methods per class.

maltoolbox/utils.py
nkakouros: Done.

maltoolbox/attackgraph/attackgraph.py
nkakouros: > Is the deepcopy stuff relevant to the LookupDicts btw?

How do you mean? They will be copied fine, like dicts. Even the indices should be copied. I don't know if we want that or not.

maltoolbox/attackgraph/attackgraph.py
mrkickling: I am ok with removing the get_by_id and get_by_full_name, but we should still have one method in attack graph which uses the LookupDict so users don't have to interact with it directly.

> As long as the user understands what fetch() or lookup() do, they won't have difficulty in reading the code

Perhaps, but I think having a method that wraps around nodes.fetch() called something like `attack_graph.get_nodes_by_attribute_value(attribute: str, value: str)` would be easier to find/understand.  And from the AttackGraphs perspective, that is what the LookupDict.fetch does.

Just to illustrate my thinking:

From the users perspective if we have no method in AttackGraph:
- They have an AttackGraph and want to fetch a node with a certain full_name
- They might look for a method in the attack graph that does this
- They will not find it, so they will think that they need to implement something themselves
- They look at attackgraph.nodes and realize it is a LookupDict
- They try to understand how the LookupDict works and find the .fetch()
- They use the .fetch('full_name', x)

If we instead had a method `attack_graph.get_nodes_by_attribute_value(attribute: str, value: str)` in the attack_graph (which uses the LookupDict.fetch), they would find it more easily.

I am skeptical of making the API less explicit by having common methods in instance variables (.nodes).

maltoolbox/attackgraph/attackgraph.py
mrkickling: What I meant was if the changes to deepcopy was relevant to the rest of the PR, but I realize now that it needed to be change because .nodes is not a dict anymore (but a LookupDict).

maltoolbox/attackgraph/attackgraph.py
nkakouros: This is where documentation helps, with examples how you do things. And IDE completion is also useful here, it will show a fetch or lookup method with full docstrings and type hints. 

I do not favor adding such a method. This code:

```
get_node_by_attribute_value(attribute, value)
```

I think does not add much compared to:

```
nodes.fetch(attribute, value)
```
especially once you have seen, read and used it once.

`nodes` is `LookupDict` and it supports an interface. If it were a dict it would support a different one (a subset). Do you think of adding wrappers around the dict.get() method or the dict.remove() method in that case? If not, why do it now? It is a custom API that not everyone is familiar with but so is the rest of the codebase and people have to study it first. It's not some hard piece of code, just that you use `nodes.fetch` to get by an attribute. 

Also, adding a get_ method would provide users with two ways to do the same thing which is not ideal.

A further motivation for doing this PR was that this LookupDict will be used elsewhere too, not just for nodes. Nodes is just the first application of this. For example, in model we have `get_asset_by_id` and `get_asset_by_id`. These can be removed both just by using a LookupDict. The interface will be one the users already now.

maltoolbox/attackgraph/attackgraph.py
nkakouros: OK, I see. We are good here then?

maltoolbox/utils.py
nkakouros: Fixed.

maltoolbox/attackgraph/attackgraph.py
mrkickling: Yes, I would add wrappers around the dicts if they were dicts. That is how it worked before with get_node_by_id and get_node_by_full_name works like this I believe (the dicts were private variables of the AttackGraph).

I agree the LookupDict could be used in more places. But I mean that in the AttackGraph it has only one use, which is to fetch nodes by attribute values / key values, which is the context the method (name) would add.

But I don't seem to be able to convince you, so maybe I will just back off.

nkakouros: We have a different programming style. I believe the changes will increase the quality of the python code, but that may not be the case at all. As a maintainer, you have a greater say, but I would hate to write back that kind of code. Maybe you want to add these changes or maybe @andrewbwm can help break the deadlock? But many of the changes I propose are of this kind, removing things that I find verbose and not pythonic and accepting some but not others will lead to an inconsistent style in the code. My mistake is that I didn't sync with you first before embarking on this. I was driven by a desire to make the code run faster and to do that it helps to have it simplified and more "linear".

mrkickling: Yes, that is it. I don't think it is easy to say who is right or wrong, we just prioritize different things. Yes, maybe @andrewbwm can chime in to give some third perspective!

nkakouros: @andrewbwm 🔔 or 🔕 ?

PR #106, opened 2025-01-28
Use fieldname as key for Language Graph Asset association dictionary

andrewbwm:
Use the fieldname relevant to the asset as the key for the associations dictionary of LanguageGraphAassets.

The idea is to use the most logical connection point rather than artificial keys we have to keep constructing, disaggregating, and reconstructing.

### DIFF
--- a/maltoolbox/language/languagegraph.py
+++ b/maltoolbox/language/languagegraph.py
@@ -60,8 +60,8 @@ class LanguageGraphAsset:
             'is_abstract': self.is_abstract
         }
 
-        for assoc in self.own_associations.values():
-            node_dict['associations'][assoc.full_name] = assoc.to_dict()
+        for fieldname, assoc in self.own_associations.items():
+            node_dict['associations'][fieldname] = assoc.to_dict()
         for attack_step in self.attack_steps.values():
             node_dict['attack_steps'][attack_step.name] = \
                 attack_step.to_dict()
@@ -317,35 +317,6 @@ class LanguageGraphAssociation:
         raise LanguageGraphAssociationError(msg % (fieldname, self.name))
 
 
-    def get_opposite_asset(
-            self, asset: LanguageGraphAsset
-        ) -> Optional[LanguageGraphAsset]:
-        """
-        Return the opposite asset if the association matches the asset given
-        as a parameter. A match can either be an explicit one or if the asset
-        given subassets either of the two assets that are part of the
-        association.
-
-        Arguments:
-        asset       - the asset to look for
-        Return the other asset if the parameter matched either of the
-        two. None, otherwise.
-        """
-        #TODO Should check to see which one is the tightest fit for
-        #     associations between assets on different levels of the same
-        #     inheritance chain.
-        if asset.is_subasset_of(self.left_field.asset):
-            return self.right_field.asset
-        if asset.is_subasset_of(self.right_field.asset):
-            return self.left_field.asset
-
-        logger.warning(
-            'Requested asset "%s" from association %s'
-            'which did not contain it!', asset.name, self.name
-        )
-        return None
-
-
 @dataclass
 class LanguageGraphAttackStep:
     name: str
@@ -484,7 +455,7 @@ class ExpressionsChain:
                     )
 
                 return {
-                    self.association.full_name:
+                    self.association.name:
                     {
                         'fieldname': self.fieldname,
                         'asset type': asset_type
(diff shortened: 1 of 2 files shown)


maltoolbox/language/languagegraph.py
```diff
@@ -567,10 +538,26 @@ def _from_dict(cls,
...
+                            assoc.name == assoc_name:
+                        association = assoc
+                        break
+
+                if association is None:
+                    msg = 'Failed to find association "%s" with '\
+                        'fieldname "%s"'
+                    logger.error(msg % (assoc_name, fieldname))
+                    raise LanguageGraphException(msg % (assoc_name,
+                        fieldname))
```
nkakouros, line 555:
```suggestion
                for assoc in target_asset.associations.values():
                    if assoc.contains_fieldname(fieldname) and \
                            assoc.name == assoc_name:
                        association = assoc
                        break
                else:
                    msg = 'Failed to find association "%s" with '\
                        'fieldname "%s"'
                    logger.error(msg % (assoc_name, fieldname))
                    raise LanguageGraphException(msg % (assoc_name,
                        fieldname))
```

maltoolbox/translators/securicad.py
```diff
@@ -141,11 +141,14 @@ def load_model_from_scad_archive(
...
-            left_field,
-            right_field,
-            left_asset.type,
-            right_asset.type)
+        lang_graph_assoc = None
+        for assoc in left_asset.lg_asset.associations:
+            if (assoc.left_field.fieldname == left_field and
+                    assoc.right_field.fieldname == right_field) or \
+                    (assoc.left_field.fieldname == right_field and
+                    assoc.right_field.fieldname == left_field):
```
nkakouros, line 149:
```suggestion
                    assoc.left_field.fieldname == right_field and
                    assoc.right_field.fieldname == left_field:
```

maltoolbox/translators/securicad.py
```diff
@@ -141,11 +141,14 @@ def load_model_from_scad_archive(
...
+                lang_graph_assoc = assoc
+                break
```
nkakouros, line 151: The `for-break-else` construct can be used here too to shave off 2 3 lines.

nkakouros:
Screenshot 2025-01-28 at 14 49 17

That's always an improvement!

maltoolbox/language/languagegraph.py
```diff
@@ -567,10 +538,26 @@ def _from_dict(cls,
                 assoc_name = list(serialized_expr_chain.keys())[0]
                 target_asset = lang_graph.assets[\
                     serialized_expr_chain[assoc_name]['asset type']]
+                fieldname = serialized_expr_chain[assoc_name]['fieldname']
```
mrkickling, line 541: Do we have a test that saves and then loads a language graph?

maltoolbox/translators/securicad.py
andrewbwm: What is the benefit of this? Are we charged per bracket?

maltoolbox/language/languagegraph.py
andrewbwm: I do not like the for-break-else format.

maltoolbox/translators/securicad.py
nkakouros: It just adds nothing. It makes the reader (me) think, wait, is there sth I didn't catch here? It also makes the file 2 bytes larger :-D

maltoolbox/language/languagegraph.py
```diff
@@ -567,10 +538,26 @@ def _from_dict(cls,
...
+                        fieldname))
+
                 new_expr_chain = ExpressionsChain(
                     type = 'field',
-                    association = target_asset.associations[assoc_name],
-                    fieldname = serialized_expr_chain[assoc_name]['fieldname']
+                    association = association,
+                    fieldname = fieldname
                 )
                 return new_expr_chain
```
mrkickling, line 562: I am a bit confused, is a languagegraph actually an expressions chain? There is not class named LanguageGraph, but it seems like ExpressionsChain is what is returned from this method that claims to return a lang graph.

andrewbwm: I think you did get confused, this is the `_from_dict` method of the `ExpressionsChain` class. Does that clear it up?

mrkickling: Does it not say for this method:
```
        """Create LanguageGraph from dict
        Args:
        serialized_expr_chain   - expressions chain in dict format
        lang_graph              - the LanguageGraph that contains the assets,
                                  associations, and attack steps relevant for
                                  the expressions chain
        """
```
Or is my github gui just confusing me..

mrkickling: No I think the docstring is probably just wrong.

PR #108, opened 2025-01-31
Merge wrappers module into attackgraph

nkakouros:


mrkickling:
Good. Before merge look for references to 'create_attack_graph' in the docs and make sure that if there are references to it they should change where they import the function from.

nkakouros:
@mrkickling I updated the docs.

A suggestion would be to have these docs in the module files themselves at the top of e.g. `attackgraph/__init__.py`. This is fairly common and makes sure that the docs are both visible to the reader of the code and that they will most likely not be skipped when updating/renaming/.. stuff.

mrkickling (approved):

nkakouros:
@mrkickling what do you think of the above?

andrewbwm (approved):

PR #109, opened 2025-02-01
Use yaml config

nkakouros:
This PR:

- simplifies the code in `maltoolbox/__init__.py`
- uses a yaml config format for consistency as the format of choice in the project
- removes the default conf file from the maltoolbox package as changing it to fit one's preferences would result in git picking it up as changed
- removed unused config options.

It changes nothing for those that did not use a custom config file. The rest will need to move to a yml config file.


maltoolbox/__init__.py
```diff
@@ -20,75 +20,56 @@
...
-import configparser
+import yaml
 import logging
+from typing import Any
+
+config: dict[str, Any] = {
+    "logging": {
+        "log_level": logging.INFO,
+        "log_file": "tmp/log.txt",
+        "attackgraph_file": "tmp/attackgraph.yml",
```
mrkickling, line 46: Is attack graph model and langspec really related to logging?

mrkickling:
I had one other thought, what if we would have a default yml file instead of having the default values in a dictionary? Does it feel better or worse? Not sure.

maltoolbox/__init__.py
nkakouros: They are, I guess, in the sense that this is where the model and attack graph are dumped for debugging.

nkakouros:
> I had one other thought, what if we would have a default yml file instead of having the default values in a dictionary? Does it feel better or worse? Not sure. But it would clarify the format of the file.

The issue with this is that there is no default path for the config file. Users will need to always create a config file even if it is the default. Unless we bundle the default config file in the maltoolbox package as it were before so that if the user installs it we know where to look for it. 

We could have a yml doc as a string instead of having the dict if that makes sense. OR, actually, we can have the default config as a resource file in the package, e.g. maltoolbox/resources/config.yml. Personally I would leave it as is, but I don't have a strong opinion over this.

maltoolbox/__init__.py
mrkickling: Yes, I think it makes sense. Now that we are at it, maybe tmp is not a good default name. What do you think about "out/" or "logs/" instead?

nkakouros: Done.

nkakouros:
Can we merge?

mrkickling:
Sorry, but in that case I think we should have in the README what the config file would look like in yml format and its default/possible values so it is easier for users to create their own config file.

Like:

## Example `maltoolbox.yml` file:

```yml
logging:
    log_level: logging.INFO # This maybe should be a string?
        log_file: logs/log.txt
        attackgraph_file: logs/attackgraph.yml
        model_file: logs/model.yml
        langspec_file: logs/langspec_file.
    neo4j:
        uri: None
        username: None
        password: None
        dbname: None
``

In my opinion we can merge it, but @andrewbwm should probably look as well.

File: tests/attackgraph/test_attackgraph.py [hunk 2]
```diff
@@ -759,7 +759,7 @@ def test_attackgraph_transitive_advanced():
 
     test_lang_graph = LanguageGraph(MalCompiler().compile(
         'tests/testdata/transitive_advanced.mal'))
-    test_lang_graph.save_to_file('tmp/trans_adv_lang_graph.yml')
+    test_lang_graph.save_to_file('logs/trans_adv_lang_graph.yml')
```
andrewbwm, line 762: Why was this changed? Was it just for testing and it was left over?

andrewbwm: Ok, I saw that this might be related to renaming the `tmp` folder to `logs` which is a great idea. But either all of the tests should save it to `logs` , or more likely this should have been saved in `/tmp/` instead of `tmp`. I'll have a look at it and fix this.

nkakouros:
> Sorry, but in that case I think we should have in the README what the config file would look like in yml format and its default/possible values so it is easier for users to create their own config file.
> 
> Like:
> ## Example `maltoolbox.yml` file:
> 
> ```yaml
> logging:
>     log_level: logging.INFO # This maybe should be a string?
>     log_file: logs/log.txt
>     attackgraph_file: logs/attackgraph.yml
>     model_file: logs/model.yml
>     langspec_file: logs/langspec_file.
> neo4j:
>     uri: None
>     username: None
>     password: None
>     dbname: None
> ```
> 
> In my opinion we can merge it, but @andrewbwm should probably look as well.

The problem with this is that you will need to remember to update this every time anyone touches the configuration. The README already points to the code with the config defaults. We could convert it to a yaml string so that it is a yaml users can copy paste. Or maybe it is worth going the extra step and adding it as a resource file.

andrewbwm:
> The problem with this is that you will need to remember to update this every time anyone touches the configuration. The README already points to the code with the config defaults. We could convert it to a yaml string so that it is a yaml users can copy paste. Or maybe it is worth going the extra step and adding it as a resource file.

I've merged this branch because you can make documentation/README changes on main regarding it if you want to.

PR #113, opened 2025-02-04
Fix "include" for some cases

nkakouros:
E.g. when doing `include ../other-lang/main.mal`.


nkakouros:
It would be nice to have some test for this but I don't feel confident touching the tests. @mrkickling Could you maybe help?

PR #118, opened 2025-02-12
Ensure that the important classes in Attack Graph, Language Graph, and Model modules have hash and repr methods.

andrewbwm:
This PR removes the dataclass decorator from two important classes `AttackGraphNode` and `Attacker`.

Another major change is that the `AttackGraphNode` class now requires the id, it is no longer optional. Which in turn means that `AttackGraphNodes` are now created as they are added to the `AttackGraph`, analogously to how the `ModelAssets` are created as they are added to a `Model`.

### DIFF
--- a/maltoolbox/attackgraph/attackgraph.py
+++ b/maltoolbox/attackgraph/attackgraph.py
@@ -204,13 +204,15 @@ class AttackGraph():
                     node_dict['lang_graph_attack_step'])
             lg_attack_step = lang_graph.assets[lg_asset_name].\
                 attack_steps[lg_attack_step_name]
-            ag_node = AttackGraphNode(
-                type=node_dict['type'],
-                lang_graph_attack_step = lg_attack_step,
-                name=node_dict['name'],
-                ttc=node_dict['ttc'],
-                asset=node_asset
+            ag_node = attack_graph.add_node(
+                lg_attack_step = lg_attack_step,
+                node_id = node_dict['id'],
+                model_asset = node_asset,
             )
+            ag_node.defense_status = node_dict.get('defense_status', None)
+            ag_node.existence_status = node_dict.get('existence_status', None)
+            ag_node.tags = set(node_dict.get('tags', {}))
+            ag_node.extras = node_dict.get('extras', {})
 
             if node_asset:
                 # Add AttackGraphNode to attack_step_nodes of asset
@@ -221,20 +223,6 @@ class AttackGraph():
                 else:
                     node_asset.attack_step_nodes = [ag_node]
 
-            ag_node.defense_status = float(node_dict['defense_status']) if \
-                'defense_status' in node_dict else None
-            ag_node.existence_status = node_dict['existence_status'] \
-                == 'True' if 'existence_status' in node_dict else None
-            ag_node.is_viable = node_dict['is_viable'] == 'True' if \
-                'is_viable' in node_dict else True
-            ag_node.is_necessary = node_dict['is_necessary'] == 'True' if \
-                'is_necessary' in node_dict else True
-            ag_node.tags = set(node_dict['tags']) if \
-                'tags' in node_dict else set()
-            ag_node.extras = node_dict.get('extras', {})
-
-            # Add AttackGraphNode to AttackGraph
-            attack_graph.add_node(ag_node, node_id=node_dict['id'])
 
         # Re-establish links between nodes.
         for node_dict in serialized_attack_steps.values():
@@ -252,7 +240,7 @@ class AttackGraph():
                                ' when loading from attack graph from dict')
                         logger.error(msg, child_id)
                         raise LookupError(msg % child_id)
-                    _ag_node.children.append(child)
+                    _ag_node.children.add(child)
 
                 for parent_id in node_dict['parents']:
                     parent = attack_graph.get_node_by_id(int(parent_id))
@@ -261,7 +249,7 @@ class AttackGraph():
                                'when loading from attack graph from dict')
                         logger.error(msg, parent_id)
                         raise LookupError(msg % parent_id)
-                    _ag_node.parents.append(parent)
+                    _ag_node.parents.add(parent)
 
         for attacker in serialized_attackers.values():
             ag_attacker = Attacker(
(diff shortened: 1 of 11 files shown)


maltoolbox/attackgraph/attackgraph.py
```diff
@@ -204,13 +205,15 @@ def _from_dict(
                     node_dict['lang_graph_attack_step'])
             lg_attack_step = lang_graph.assets[lg_asset_name].\
                 attack_steps[lg_attack_step_name]
-            ag_node = AttackGraphNode(
-                type=node_dict['type'],
-                lang_graph_attack_step = lg_attack_step,
-                name=node_dict['name'],
-                ttc=node_dict['ttc'],
-                asset=node_asset
+            ag_node = attack_graph.add_node(
```
mrkickling, line 208: add_node could take optional `defense_status`, `existence_status` and maybe `tags`, `extras`.

andrewbwm: The defense status and existence status I can be swayed on. I don't like the idea of having `tags` in there as the base value is set based on the language graph attack step and `extras` should primarily be handled entirely outside of the mal-toolbox. I think the toolbox should just create the dictionary as a hint of the place where such values should be stored, but do nothing else with them apart from saving and restoring them to/from a file.

mrkickling: Sounds good to me. Otherwise I approve this to merged as soon as possible.

maltoolbox/language/languagegraph.py
```diff
@@ -104,7 +105,7 @@ def to_dict(self) -> dict:
 
 
     def __repr__(self) -> str:
-        return str(self.to_dict())
+        return 'LanguageGraphAsset(name: "%s")' % self.name
```
nkakouros, line 108: It's a minor thing, but `%`-formatting is considered clunky, old and less preferred that f-strings and `format`. Maybe also for consistency it makes sense to stick to one formatting style e.g. f-strings and use `format` as a backup in cases where f-strings are not suitable?

maltoolbox/attackgraph/node.py
```diff
@@ -89,28 +97,20 @@ def __deepcopy__(self, memo) -> AttackGraphNode:
...
+            node_id = self.id,
+            model_asset = self.model_asset,
+            lg_attack_step = self.lg_attack_step
         )
 
         copied_node.tags = copy.deepcopy(self.tags, memo)
-        copied_node.attributes = copy.deepcopy(self.attributes, memo)
         copied_node.extras = copy.deepcopy(self.extras, memo)
+        copied_node.ttc = copy.deepcopy(self.ttc, memo)
+        copied_node.detectors = copy.deepcopy(self.detectors, memo)
```
nkakouros, line 108: These  (except `extras`) should not change after the node is created and they will already be set during node creation (in __init__) right? Is there an explicit need to copy them over here?

maltoolbox/attackgraph/attackgraph.py
```diff
@@ -715,31 +695,54 @@ def regenerate_graph(self) -> None:
...
+            node_id = node_id,
+            lg_attack_step = lg_attack_step,
+            model_asset = model_asset
+        )
 
         self.nodes.append(node)
         self._id_to_node[node.id] = node
         self._full_name_to_node[node.full_name] = node
 
+        return node
```
nkakouros, line 744: I think we can do away with add_node by having the AttackGraphNode class track the maximum node id. This would be a global counter, meaning that it will be shared among different attack graph if more than one are created during a python run. We don't do that anywhere, but maybe in the future it happens? In that case, the user that creates these graphs should reset AttackGraphNode.id to 0 before starting building the next graph.

What I am proposing is sth like:

```
# when building the attack graph
node = AttackGraphNode(....)
#OR
node = AttackGraphNode(id=1234)
attackgraph.nodes.append(node)
# merging the lookupdict PR means `node` does not need to be appended elsewhere


# Then in node.py we do:
class AttackGraphNode:
    _max_id = -1

    def __init__(self, id=None, ....):
        ...
        AttackGraphNode._max_id = max(self._id + 1, id or 0)
        self.id = AttackGraphNode._max_id
        ....
```

So, with 3 lines we can replace ~ 45 lines. `_max_id` above is bound to the class. The user can do AttackGraphNode._max_id = -1 to start building a second attack graph with nodes starting from 0 (or they may not do that and just add nodes to the second graph with ids starting from whatever _max_id ended up being after the first graph was created). We may also provide a static method `AttackGraphNode.reset()` or sth to make this more user-friendly.

maltoolbox/attackgraph/attackgraph.py
```diff
@@ -715,31 +695,54 @@ def regenerate_graph(self) -> None:
...
 
-        node.id = node_id if node_id is not None else self.next_node_id
-        self.next_node_id = max(node.id + 1, self.next_node_id)
+        node = AttackGraphNode(
+            node_id = node_id,
+            lg_attack_step = lg_attack_step,
+            model_asset = model_asset
+        )
 
         self.nodes.append(node)
```
nkakouros, line 740: BTW, can't we make the node list in the attack graph be a set? Will this break anything? Conceptually, nodes in a graph are unordered. Also, a node should only appear once given that each node has a compromised_by list. This means that using a list which keeps the order of items and allows for multiplicity is not needed.

maltoolbox/attackgraph/node.py
andrewbwm: Good catch, some of the code changed and I forgot to update it. I will still copy over the `tags` and `ttc`. I think we should assume that they could be modified from the language default, but the others can be skipped.

maltoolbox/attackgraph/attackgraph.py
andrewbwm: We are definitely not doing this in this pull request. I'm ok with the way they are now, but am also ok with changing them. This isn't at all a priority though right now.

maltoolbox/language/languagegraph.py
andrewbwm: I will at one point lock all of you in a room and when only sole survivor remains I will just take their advice. I don't remember whom it was that suggested that we switch over to the c-style strings in the debug logs.

maltoolbox/attackgraph/attackgraph.py
nkakouros: OK. I will create a PR for when there is time to review.

maltoolbox/language/languagegraph.py
andrewbwm: I think the issue is that you can't do the comma one for strings outside of the methods that know how to handle variable numbers of string parameters. It doesn't matter, I will just add this to the code cleanup issue and we will fix it when we get to it in 2077.

mrkickling: We should still use c-style in log statements (for performance). For other cases we might use f-strings if it looks better since it does not affect performance either way.

maltoolbox/attackgraph/attackgraph.py
andrewbwm: Better still, we already had id keyed dictionaries both both the nodes and the attackers.

maltoolbox/language/languagegraph.py
nkakouros: > We should still use c-style in log statements (for performance). For other cases we might use f-strings if it looks better since it does not affect performance either way.

The important distinction is that arguments should not be given after a `%` but as arguments to the logging function.

mrkickling: Yes, maybe c-style is the wrong term, not very used to c.

maltoolbox/attackgraph/attackgraph.py
nkakouros: https://github.com/mal-lang/mal-toolbox/pull/119

Issue #98, opened 2025-01-17
Enable performance tracking in CI


nkakouros:
Use the biggest model we have and use that to check for performance regressions in PRs.

mrkickling:
Sounds good. Question is though where it should be displayed/stored. Maybe there is some common way to do it in the github pipeline..

nkakouros:
There are websites where Github actions can send results and which will keep a history of performance. And they will send an error back to GitHub if performance is degraded. This is similar to how coverage can be calculated and checked in GitHub actions (which is also a good idea to enable in CI).

PR #121, opened 2025-02-17
Various changes for malsim PR #98

nkakouros:
See the commits for each change. Regardless of malsim PR #98, I believe these are good to have changes.

### DIFF
--- a/maltoolbox/attackgraph/query.py
+++ b/maltoolbox/attackgraph/query.py
@@ -114,10 +114,13 @@ def is_node_traversable_by_attacker(
             )
             return False
 
-def get_attack_surface(
+
+def calculate_attack_surface(
         attacker: Attacker,
-        nodes: Iterable[AttackGraphNode] | None = None,
-    ) -> set[AttackGraphNode]:
+        *,
+        from_nodes: Iterable[AttackGraphNode] | None = None,
+        skip_compromised: bool = False,
+) -> set[AttackGraphNode]:
     """
     Get the current attack surface of an attacker. This includes all of the
     viable children nodes of already reached attack steps that are of 'or'
@@ -125,9 +128,10 @@ def get_attack_surface(
     parents in the attack steps reached.
 
     Arguments:
-    attacker    - the Attacker whose attack surface is sought
-    nodes       - the nodes to calculate the attack surface from; defaults to
-                  the attackers compromised nodes list if omitted
+    attacker          - the Attacker whose attack surface is sought
+    nodes             - the nodes to calculate the attack surface from; defaults
+                        to the attackers compromised nodes list if omitted
+    skip_compromised  - do not add already compromised nodes to the attack surface
     """
     logger.debug(
         'Get the attack surface for Attacker "%s"(%d).',
@@ -135,7 +139,7 @@ def get_attack_surface(
         attacker.id
     )
     attack_surface = set()
-    for attack_step in nodes or attacker.reached_attack_steps:
+    for attack_step in from_nodes or attacker.reached_attack_steps:
         logger.debug(
             'Determine attack surface stemming from '
             '"%s"(%d) for Attacker "%s"(%d).',
@@ -145,6 +149,8 @@ def get_attack_surface(
             attacker.id
         )
         for child in attack_step.children:
+            if skip_compromised and child.is_compromised_by(attacker):
+                continue
             if is_node_traversable_by_attacker(child, attacker) and \
                     child not in attack_surface:
                 logger.debug(
--- a/maltoolbox/attackgraph/query.py
+++ b/maltoolbox/attackgraph/query.py
@@ -7,7 +7,7 @@ attack graph, but do not alter the structure or nodes in any way.
 from __future__ import annotations
 from collections.abc import Iterable
 import logging
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Optional
 
 from .attackgraph import AttackGraph, Attacker
 
@@ -118,24 +118,25 @@ def is_node_traversable_by_attacker(
 def calculate_attack_surface(
         attacker: Attacker,
         *,
-        from_nodes: Iterable[AttackGraphNode] | None = None,
+        from_nodes: Optional[Iterable[AttackGraphNode]] = None,
         skip_compromised: bool = False,
 ) -> set[AttackGraphNode]:
     """
-    Calculate the attack surface for the attacker. If from_nodes is provided,
-    calculate attack surface as if the attacker was standing on those nodes,
-    otherwise use all nodes the attacker has compromised. If skip_compromised is
-    set, exclude already compromised nodes from the returned attack surface.
+    Calculate the attack surface of the attacker. If from_nodes are provided
+    only calculate the attack surface stemming from those nodes, otherwise use
+    all nodes the attacker has compromised. If skip_compromised is true,
+    exclude already compromised nodes from the returned attack surface.
 
-    The attack surface includes all of the viable children nodes of from_nodes
-    that are of 'or' type and the 'and' type children nodes which have all of
-    their necessary parents compromised but the attacker.
+    The attack surface includes all of the viable children nodes that are of
+    'or' type and the 'and' type children nodes which have all of their
+    necessary parents compromised by the attacker.
 
     Arguments:
     attacker          - the Attacker whose attack surface is sought
-    nodes             - the nodes to calculate the attack surface from; defaults
+    from_nodes        - the nodes to calculate the attack surface from; defaults
                         to the attackers compromised nodes list if omitted
-    skip_compromised  - do not add already compromised nodes to the attack surface
+    skip_compromised  - if true do not add already compromised nodes to the
+                        attack surface
     """
     logger.debug(
         'Get the attack surface for Attacker "%s"(%d).',
@@ -158,7 +159,7 @@ def calculate_attack_surface(
             if is_node_traversable_by_attacker(child, attacker) and \
                     child not in attack_surface:
                 logger.debug(
-                    'Add node "%s"(%d) to the attack surface for '
+                    'Add node "%s"(%d) to the attack surface of '
                     'Attacker "%s"(%d).',
                     child.full_name,
                     child.id,
--- a/maltoolbox/attackgraph/query.py
+++ b/maltoolbox/attackgraph/query.py
@@ -122,10 +122,14 @@ def calculate_attack_surface(
         skip_compromised: bool = False,
 ) -> set[AttackGraphNode]:
     """
-    Get the current attack surface of an attacker. This includes all of the
-    viable children nodes of already reached attack steps that are of 'or'
-    type and the 'and' type children nodes which have all of their necessary
-    parents in the attack steps reached.
+    Calculate the attack surface for the attacker. If from_nodes is provided,
+    calculate attack surface as if the attacker was standing on those nodes,
+    otherwise use all nodes the attacker has compromised. If skip_compromised is
+    set, exclude already compromised nodes from the returned attack surface.
+
+    The attack surface includes all of the viable children nodes of from_nodes
+    that are of 'or' type and the 'and' type children nodes which have all of
+    their necessary parents compromised but the attacker.
 
     Arguments:
     attacker          - the Attacker whose attack surface is sought
--- a/maltoolbox/attackgraph/query.py
+++ b/maltoolbox/attackgraph/query.py
@@ -5,6 +5,7 @@ This submodule contains functions that analyze the information present in the
 attack graph, but do not alter the structure or nodes in any way.
 """
 from __future__ import annotations
+from collections.abc import Iterable
 import logging
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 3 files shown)


maltoolbox/attackgraph/query.py
```diff
@@ -111,14 +113,16 @@ def get_attack_surface(
 
     Arguments:
     attacker    - the Attacker whose attack surface is sought
+    nodes       - the nodes to calculate the attack surface from; defaults to
```
mrkickling, line 116: Could you explain the idea behind `nodes`?
If you give `nodes`, the attacker should becomes irrelevant it feels like?

In that case, create a new function called `get_attack_surface_from_nodes` that only takes a list of nodes and have `get_attack_surface(attacker)` just return `get_attack_surface_from_nodes(attacker.reached_attack_steps)`.

If you call `get_attack_surface(attacker, nodes)`, and attacker does not have `nodes` in its reached_attack_steps, how can they even be `traversable` according to `is_node_traversable_by_attacker`?

mrkickling:
Other than my comments/questions about the `get_attack_surface`, converting various return values to sets to show that they are not ordered makes sense.

maltoolbox/attackgraph/query.py
nkakouros: I wanted to remove `update_attack_surface_add_nodes` because it did 99% of what `get_attack_surface` did and it also mutated the attacker's `attack_surface` property, despite also returning the updated attack surface, which was out of the blue. So, I merged these two together. 

I agree that if nodes given the attacker arg is not used. This means we can either:

- combine nodes and attacker args into one with a type hint of `Attacker | list[AttackGraphNode]`
- or, remove the attacker arg and just use nodes. Then they way you would call it is:
```
get_attack_surface(attacker.reached_attack_steps)
```

I think the latter is preferable because it allows you to limit the nodes from where to search for new attack surface. Below you can see the only two places where this function is used:

```
malsim/sims/mal_simulator.py:112:                agent.action_surface = query.get_attack_surface(attacker)
malsim/sims/mal_simulator.py:249:        new_attack_surface = query.get_attack_surface(attacker, actioned_nodes)
```

The first is when the simulation starts and `reached_attack_steps` will contain only the entry points. The second is how I have changed it to look like. Before it would loop through aaaaaall reached attack steps without needing too. Then, I combine the `new_attack_surface` with the old:

```
malsim/sims/mal_simulator.py:249:        new_attack_surface = query.get_attack_surface(attacker, actioned_nodes)
malsim/sims/mal_simulator.py-250-        # TODO: why extend?
malsim/sims/mal_simulator.py-251-        agent.action_surface |= new_attack_surface
```

The old will have already captured any attack surface that the older nodes in `reached_attack_steps` have allowed. This will make it go faster. Also, the function now has more utility as it covers more cases.

The above will keep working even if a defense action makes some nodes in the attacker's old action_surface unviable because whenever a defense is enabled, each attacker's attack_surface gets searched and emptied of such now unviable steps.

The only way this will fail is if a defense is disabled and some nodes become viable again, in which case all the attack surface and not just the new additions to it will need to be examined if they allow new child steps. But this is a use not supported anywhere in the code as far as I can tell. Even if it is supported it is easy to use the new `get_attack_surface` and just pass it the whole attack surface instead of just the new additions to the attack surface.

One thing you may object is the naming of the function, `get_attack_surface`. You may think that it "requires" an attacker, reading it as `get the attack surface for the attacker`. But the way I read it is `get the attack surface reachable from these nodes`. My reading views attack surface as a system property and not an attacker one, which I think is more fitting in this case since this function exists in the mal-toolbox which has little to no concept of attackers.

nkakouros: > If you call `get_attack_surface(attacker, nodes)`, and attacker does not have `nodes` in its reached_attack_steps, how can they even be `traversable` according to `is_node_traversable_by_attacker`?

I read your comment and the code more carefully and I realize that as it stands now the attacker cannot be removed. What you mention is fine, the node will not be traversable and no attack surface will be found from that node for that attacker. I don't see why this contrasts anything. 


> In that case, create a new function called `get_attack_surface_from_nodes` that only takes a list of nodes and have `get_attack_surface(attacker)` just return `get_attack_surface_from_nodes(attacker.reached_attack_steps)`.

If you have two functions:

- `get_attack_surface`
- `get_attack_surface_from_nodes`

doesn't it create the question to the first time user what the first function gets the attack surface from? You will have to rename it to `get_attack_surface_for_attacker` but that still does not answer that. You can name it `get_attack_surface_from_compromised_nodes` which is way too long and more importantly makes you wonder, why can't I combine this with the other method that does the same thing but for any set of nodes and not just the compromised one? Or they may think , wait, the other method also gets attack surface from nodes, what is going on? Generally, this is a sign that you want to refactor sth to simplify it (which will lead you to sth like what I propose in this PR). You can argue that if the user reads the doc for `get_attack_surface` they will understand what it does. But so will they if they read the doc of the updated `get_attack_surface` I provide here.

Generally, and I mean it in a constructive way, adding functions with long names doing sth a little bit differently than others with similar names with maybe different return signatures, and and and, is bad. It makes the code harder to read, harder to maintain. Doing one thing and doing it well is important, but if that "one thing" can be abstracted away more without breaking semantics, making the code hard, etc, it's even better.

nkakouros: Sth more to add, I plan at some point a PR that abstracts away the maltoolbox.attackgraph.Attacker, basically removing that class. If you look at that class it all boils down to 4 lines of code:

```
node.compromised_by.add(self)
self.reached_attack_steps.add(node)

# and the opposite
node.compromised_by.remove(self)
self.reached_attack_steps.remove(node)
```

These are node operations that can happen already on the node objects. For example, via a `node.compromise()` (this is just an example I have not thought through this example). This will generalize the `compromise` concept (e.g. how do you go about now if you want an agent that is both a defender and an attacker, like an insider threat?) and will make the mal simulator code simpler. Currently, in the mal simulator there are agents from the malsim.agents subpackage, then there is the agent views which at places are also called agents and then there is the attacker from the toolbox. This is messy to read in the code I think.

I am saying the above because having the attacker be an arg to `get_attack_surface` is not strictly needed based on what conceptually happens, but the current code makes it necessary to be there.

nkakouros: Do we have a consensus here or do we need to discuss/change more?

mrkickling: I will have to read through everything more carefully and test the new function a bit before we have consensus.

Just a comment on your last comment:
If we remove the Attacker concept, how do you keep track on which attacker has compromised which nodes? There has to be some sort of attacker concept, no?

nkakouros: This is for a future redesign how exactly we do this. But a raw idea is to have each node keep track of who compromised them. Or have a general-purpose registry where agents can write whatever they need, compromised status being one.

About tests, it will take me some time to put sth useful in place. If you can provide some skeleton or sth, it would be helpful.

mrkickling: You are right that I will object the naming. The function is called get_attack_surface, and in my opinion that means it should return the attack surface.
- If you give the function `attacker` but no `nodes`, it will return the attack surface.
- If you give the function `attacker` and `nodes` the function instead returns the diff of the action surface in the case that those nodes would be compromised (at least that is what I saw when trying it out in a test)

I find this confusing. Yes, I think it would be less confusing to have `get_attack_surface(attacker: Attacker)` and `get_attack_surface_from_nodes(nodes: set[AttackGraphNodes])`  (or would the second one also need the Attacker?). And of course `get_attack_surface` would use the other function if possible so we don't have code duplication.

I do agree that having long function names and many functions is not a good thing on its own, and it can be bad.  If there is duplicate code which you mentioned was the case in these functions I totally agree something should be done. But there is a dilemma: Do we want to prioritize understandable API with clear explicit function names / responsibility or do we want the code to have fewer lines and fewer functions? I believe that small functions with clear responsibility that make use of each other and have explicit names is a better approach. But this is a matter of opinion, so maybe we will not end up with a solution we are both happy with.

mrkickling: Here is a some ideas for testing, could be helpful. Uses the example_attackgraph from conf_test which is a fixture.

```
def test_attacker_attack_surface(example_attackgraph: AttackGraph):
    """Attack a node and see expected behavior"""

    calculate_viability_and_necessity(example_attackgraph)

    # Create the attacker and add with entrypoint to the graph
    app1_nw_conn_uninspected = (
        example_attackgraph.get_node_by_full_name(
            'Application 1:networkConnectUninspected'
        )
    )
    assert app1_nw_conn_uninspected
    attacker = Attacker(
        'attacker',
        entry_points={app1_nw_conn_uninspected},
        reached_attack_steps={app1_nw_conn_uninspected}
    )
    example_attackgraph.add_attacker(attacker)

    # Attacker starts from x
    expected_attack_surface = {
        'Application 1:specificAccessNetworkConnect',
        'Application 1:softwareProductVulnerabilityNetworkAccessAchieved',
        'Application 1:attemptUseVulnerability',
        'Application 1:networkConnect'
    }

    app1_attempt_use_vuln = (
        example_attackgraph.get_node_by_full_name(
            'Application 1:attemptUseVulnerability'
        )
    )
    assert app1_attempt_use_vuln

    attack_surface = get_attack_surface(attacker)
    attack_surface_full_names = {n.full_name for n in attack_surface}
    assert attack_surface_full_names == expected_attack_surface

    ### Compromise attemptUseVulnerability
    attacker.compromise(app1_attempt_use_vuln)
    attack_surface = get_attack_surface(attacker)
    attack_surface_full_names = {n.full_name for n in attack_surface}
    # Should add successfulUseVulnerability to attack surface
    expected_attack_surface |= {'Application 1:successfulUseVulnerability'}
    assert attack_surface_full_names ==  expected_attack_surface

    app1_suc_use_vuln = (
        example_attackgraph.get_node_by_full_name(
            'Application 1:successfulUseVulnerability'
        )
    )
    assert app1_suc_use_vuln

    ### Compromise successfulUseVulnerability
    attacker.compromise(app1_suc_use_vuln)
    attack_surface = get_attack_surface(attacker)
    attack_surface_full_names = {n.full_name for n in attack_surface}
    # Should add useVulnerability to attack surface
    expected_attack_surface |= {'Application 1:useVulnerability'}
    assert attack_surface_full_names == expected_attack_surface

    app1_use_vuln = (
        example_attackgraph.get_node_by_full_name(
            'Application 1:useVulnerability'
        )
    )
    assert app1_use_vuln
    attack_surface = get_attack_surface(attacker, {app1_use_vuln})
    # if we give nodes as second argumen attack surface is not returned
    # but instead the 'diff' of the attack surface if the given
    # nodes would be compromised
    assert (
        {n.full_name for n in attack_surface}
        == {'Application 1:attemptSoftwareProductAbuse'}
    )

    # Calling it without nodes gives the full attack surface
    attack_surface = get_attack_surface(attacker)
    assert {n.full_name for n in attack_surface} == expected_attack_surface

```

nkakouros: Would you settle for a single function like this:

```python
def calculate_attack_surface(
    attacker: Attacker,
    *,
    from_nodes: set[AttackGraphNode,
)
```

so that when you call it, it reads:

```
attack_surface = query.calculate_attack_surface(attacker)

# and

new_additions_to_attack_surface = query.calculate_attack_surface(attacker, just_compromised_nodes)
```

nkakouros: > If you give the function attacker and nodes the function instead returns the diff of the action surface in the case that those nodes would be compromised (at least that is what I saw when trying it out in a test)

It should return the nodes that are reachable from the passed in `nodes` list and attackable for the specific attacker. Is this what you describe?

mrkickling: > attack_surface = query.calculate_attack_surface(attacker)
>
># and
>
>new_additions_to_attack_surface = query.calculate_attack_surface(
>   attacker,
>    from_nodes = just_compromised_nodes
>)

To me the change from `nodes` to `from_node` actually made quite a big difference, so I am ok with it as long as it is documented in the docstring what happens when you give `from_nodes`.

nkakouros: I changed the function's docstring. Let's merge if you feel OK with the PR and move on to the mal sim PR.

maltoolbox/attackgraph/query.py
```diff
@@ -141,56 +153,23 @@ def get_attack_surface(
...
-            attack_step.id,
-            attacker.name,
-            attacker.id
-        )
-        for child in attack_step.children:
-            is_traversable = is_node_traversable_by_attacker(child, attacker)
-            if is_traversable and child not in attack_surface:
                 logger.debug(
-                    'Add node "%s"(%d) to the attack surface of '
+                    'Add node "%s"(%d) to the attack surface for '
```
andrewbwm, line 161: I'm perfectly happy with my English, by the way.

nkakouros: I didn't mean it as a syntax correction. I changed that to be on par with the docstring wording I had, "attack surface for", to indicate it is not THE attack surface of the attacker that you will get, it will depend on from_nodes whether it is. I.e. you will get AN attack surface for the attacker based on the arguments you supply in the function.

PR #123, opened 2025-02-21
Set Attacker defaults

nkakouros:
It does not break anything and removes redundant lines of code.


mrkickling:
Agreed to have defaults, but why was deepcopy removed?

nkakouros:
> Agreed to have defaults, but why was deepcopy removed?

Because it did what the default __deepcopy__ implementation does.

mrkickling (approved):

PR #126, opened 2025-03-07
Equality Overrides

sandorstormen:
I changed equality checks for:
- AttackGraphNode, with these properties:
  - ID
  - Name 
  - ModelAsset
- ModelAsset, with these properties
  - ID
  - Name
  - Type

### DIFF
--- a/maltoolbox/attackgraph/node.py
+++ b/maltoolbox/attackgraph/node.py
@@ -118,6 +118,15 @@ class AttackGraphNode:
 
         return copied_node
 
+    def __eq__(self, other):
+        if not isinstance(other, AttackGraphNode):
+            return False
+
+        same_id = self.id == other.id
+        same_name = self.name == other.name
+        same_asset = self.model_asset == other.model_asset
+
+        return same_id and same_name and same_asset
 
     def is_compromised(self) -> bool:
         """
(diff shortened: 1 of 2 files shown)


mrkickling:
I think the reason why you wanted this is so the simulator treats two nodes like the same node, even though they are from a different attack graph than the one in the simulator (this happens because of reset).

I think the solution should be in the simulator instead.

nkakouros:
> I changed equality checks for:
> 
>     * AttackGraphNode, with these properties:
>       
>       * ID
>       * Name
>       * ModelAsset
> 
>     * ModelAsset, with these properties
>       
>       * ID
>       * Name
>       * Type

Intuitively, there is no clear definition for when a node or asset equals another. You limit here for example node equality to be about name and model-asset. But if one node is compromised by some attacker and the other is not, are they still equal?

Including the ID does not make sense to me. If the ID is the same, then the two Nodes or Assets are the same (not just equal), or should be the same otherwise the property should not be called ID. So, the ID should be excluded from equality checks.

Unless, the equality check is used to compare only IDs and nothing else in which case it will be a weaker version of `nodeA is NodeB`; the two nodes will not be the same object in memory but they are functionally equivalent for the attack graph logic. The same for model assets.

PR #128, opened 2025-03-21
Add ttc sampling and expected value to AttackGraphNode

mrkickling:


nkakouros:
> when sampling ttc with bernoulli and bernoulli returns 0, ttc value for that sample will be infinite. Does this make sense?

Reading the [wiki](https://github.com/mal-lang/malcompiler/wiki/Supported-distribution-functions#bernoulli-behaviour):

> An attack step with a TTC function Bernoulli(p) means that the attack step can be performed immediately with a probability of p, and not at all with a probability of 1 - p.

To me, this means "disable the step". What you suggest (setting a TTC of infinite) is similar but different. The agents can still attempt the step but never progress past it. However, according to the wiki, the step should not be "performable" at all.

maltoolbox/attackgraph/node.py
```diff
@@ -181,6 +183,101 @@ def is_available_defense(self) -> bool:
...
+            raise ValueError(f"Unknown TTC distribution: {distribution}")
+
+        return s
+
+    def ttc_expected_value(self) -> float:
+        """Returns the expected value of the ttc distribution for a node.
+
+        TTC Distributions in MAL:
+        https://mal-lang.org/mal-langspec/apidocs/org.mal_lang.langspec/org/mal_lang/langspec/ttc/TtcDistribution.html
+        """
```
nkakouros, line 240: Why is this needed? Didn't we discuss in the last meeting that sampling TTC distributions would return the TTC value directly?

maltoolbox/attackgraph/node.py
```diff
@@ -181,6 +183,101 @@ def is_available_defense(self) -> bool:
             'suppress' not in self.tags and \
             self.defense_status != 1.0
 
+    def ttc_sample(self) -> float:
+        """Sample a value from ttc distribution for a node
+
+        TTC Distributions in MAL:
+        https://mal-lang.org/mal-langspec/apidocs/org.mal_lang.langspec/org/mal_lang/langspec/ttc/TtcDistribution.html
+        """
```
nkakouros, line 191: The link refers to the Gamma, Binomial and other distributions. But there are missing from the implementation here. Am I missing sth?

mrkickling:
> > when sampling ttc with bernoulli and bernoulli returns 0, ttc value for that sample will be infinite. Does this make sense?
> 
> Reading the [wiki](https://github.com/mal-lang/malcompiler/wiki/Supported-distribution-functions#bernoulli-behaviour):
> 
> > An attack step with a TTC function Bernoulli(p) means that the attack step can be performed immediately with a probability of p, and not at all with a probability of 1 - p.
> 
> To me, this means "disable the step". What you suggest (setting a TTC of infinite) is similar but different. The agents can still attempt the step but never progress past it. However, according to the wiki, the step should not be "performable" at all.

Removing the attack step completely might be the solution then. But maybe that is a choice to be taken by the simulator when it samples the ttcs and removing those that has a sampled ttc that disables the node.

maltoolbox/attackgraph/node.py
```diff
@@ -181,6 +183,101 @@ def is_available_defense(self) -> bool:
...
+
+            If the Bernoulli trial fails (0), return infinity (impossible).
+            If the Bernoulli trial succeeds (1), return sample from
+            exponential distribution.
+            """
+
+            # If bernoulli is set to 1, the sample is just exponential
+            if np.random.choice([0, 1], p=[1 - bernoulli, bernoulli]):
+                return np.random.exponential(scale=1 / exponential)
+            return math.inf
```
nkakouros, line 207: I think this is too convoluted. Why not use `np.random.exponential` directly in all cases in the lines below, except the bernoulli one where you do just random.choice? The way it is now, it seems like you are doing Bernoulli * Exponential, when the TTC in the mal spec does may not use the Bernoulli distribution at all and you have to default to the Bernoulli part returning 1 for such cases. While the result is the same, it is unnecessarily complex.

maltoolbox/attackgraph/node.py
```diff
@@ -181,6 +183,101 @@ def is_available_defense(self) -> bool:
...
+            simply be from an exponential.
+
+            If the Bernoulli trial fails (0), return infinity (impossible).
+            If the Bernoulli trial succeeds (1), return sample from
+            exponential distribution.
+            """
+
+            # If bernoulli is set to 1, the sample is just exponential
+            if np.random.choice([0, 1], p=[1 - bernoulli, bernoulli]):
+                return np.random.exponential(scale=1 / exponential)
```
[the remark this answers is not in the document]
nkakouros: Also, this will return a float. The TTC is an integer. There should be some rounding done here.

maltoolbox/attackgraph/node.py
nkakouros: After I wrote this, I realized that most of the distributions here are the codenamed ones, eg hard and uncertain which do use the Bernoulli distribution in their formula. But still others, like the pure exponential, do not and I still think that this could be simplified a bit. 

Maybe it merits an online call to discuss some aspects of this.

maltoolbox/attackgraph/node.py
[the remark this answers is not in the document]
mrkickling: Did we decide that the TTC is an integer?

maltoolbox/attackgraph/node.py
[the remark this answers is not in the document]
mrkickling: You mean putting the ttc calculations outside of the AttackGraphNode?

maltoolbox/attackgraph/node.py
mrkickling: Good catch, they are missing. I started off from a previous implementation that did not use all of the distributions. I would need help to implement those.

andrewbwm:
So, I think I'd like two things to happen.

1.  Have the toolbox just have a collection of predefined TTCs, such as `HardAndUncertain`, and when we process the language graph we simply replace it with what it actually represents(Bernouli(0.5) * Exponential(0.1) or whatever it is). We may even want to save this in a separate file for easy editing.
2. Have the entire sampling logic happen in the Simulator. TTCs are yet another ambiguous feature that kind of lives in the liminal space between toolbox and simulator. However, it really doesn't have any intrinsic meaning within the toolbox, so I'd be quite happy to just move it wholesale to the simulator. We also expect most people to use the simulator if they intend to do any sort of Attack Graph traversal at this point anyway.

tests/testdata/prob_dists.mal
```diff
@@ -0,0 +1,16 @@
+// vim:ft=mal
+#id: "test"
+#version: "0.0.0"
+
+category Test{
+  asset Dummy {
+    | expo [Exponential(0.1)]
+    | berni [Bernoulli(0.5)]
+    | composite [Exponential(0.1) * Bernoulli(0.5)]
+    | compositeComplex [Bernoulli(0.5) * (Binomial(10, 0.1) + Exponential(0.2) - Gamma(0.1, 0.2) * LogNormal(5, 0.5) / Pareto(7, 0.25) ^ TruncatedNormal(6, 0.3) + Uniform(0.1, 0.9))]
```
andrewbwm, line 10: @nkakouros just as a heads up, we don't need it right now since it will not be used any time soon, but the compiler does not handle the `/` correctly. The expression simply terminates when it encounters it.

maltoolbox/language/languagegraph.py
```diff
@@ -654,6 +654,7 @@ class LanguageGraph():
     """Graph representation of a MAL language"""
     def __init__(self, lang: Optional[dict] = None):
         self.assets: dict = {}
+        self.load_predefined_ttcs()
```
mrkickling, line 657: I don't think this needs to be a method since all it does is to set an instance variable (predef_ttcs). I would either remove the method or have the method return the predefined ttc values so the instance variable is defined in `__init__`.

maltoolbox/probs_utils.py
```diff
@@ -0,0 +1,174 @@
...
+        case 'Exponential':
+            lambd = float(probs_dict['arguments'][0])
+            return random.expovariate(lambd)
+
+        case 'Binomial':
+            n = int(probs_dict['arguments'][0])
+            p = float(probs_dict['arguments'][1])
+            # TODO: Someone with basic probabilities competences should
+            # actually check if this is correct.
+            return random.binomialvariate(n, p)
```
mrkickling, line 43: My linter tells me this method does not exist in module random

mrkickling: I am also not the correct person to say if these sampling methods are correct :P

andrewbwm: https://docs.python.org/3/library/random.html#random.binomialvariate

I think the issue is that it was introduced in python 3.12

maltoolbox/probs_utils.py
```diff
@@ -0,0 +1,174 @@
+"""Utility functions for handling probabilities"""
+
+import logging
+import math
+import random
+from enum import Enum
+
+logger = logging.getLogger(__name__)
+
+class TTCCalculation(Enum):
```
mrkickling, line 10: `TTCCalculationMethod`, `TTCMethod`, `TTCCalcMethod`.
It is not a calculation in itself, but used to specify which method to calculate the TTC with, do you agree?

mrkickling:
Looks good to me, but I can not say if the sampling methods are correct

maltoolbox/probs_utils.py
```diff
@@ -0,0 +1,174 @@
+"""Utility functions for handling probabilities"""
```
mrkickling, line 1: Maybe the module/file should actually be ttc_utils.py? At least at this point it only does ttc calculations.

maltoolbox/probs_utils.py
andrewbwm: Yes, I considered `TTCCalculationMethod` and I don't like any of the names, so I'll just switch it to that for the time being.

maltoolbox/probs_utils.py
mrkickling: Oh alright, that would make mal-toolbox require 3.12 I guess so maybe it is worth to see if there is a different option.

andrewbwm: There isn't and something somewhere(maybe in the mal simulator) already requires 3.12. Some people trying to use it ran into it. So, I think we should just specific 3.12 as a requirement.

maltoolbox/probs_utils.py
andrewbwm: Again, not true, these can be used for anything that wants a probability, I just wrote ttc all over the place because I was focusing on the implementation, I will probably rename them. But I first need to figure out what the whole structure looks like.

maltoolbox/language/predefined_ttcs.yml
```diff
@@ -0,0 +1,59 @@
+EasyAndUncertain:
```
mrkickling, line 1: I am sceptical if having a yml with the predefined ttcs is helpful. Would it not be clearer to have it as a dict in the prob_dists.mal or somewhere else? What is the argument for a yml?

maltoolbox/language/languagegraph.py
```diff
@@ -692,6 +694,30 @@ def from_mar_archive(cls, mar_archive: str) -> LanguageGraph:
...
+
+        Arguments:
+        ttc_entry   - the TTC entry to check for predefined names
+
+        Returns:
+        If the TTC entry provided contained a predefined name the TTC
+        probability distrubtion corresponding to it. Otherwise, the TTC
+        distribution provided as a parameter as is.
+        """
+        if ttc_entry is None:
```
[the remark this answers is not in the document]
andrewbwm: Many attack steps won't have ttc distributions associated with them. They could either have None or empty dictionaries as values. I do not recall right now which one we use or if both somehow are in use.

maltoolbox/language/languagegraph.py
```diff
@@ -692,6 +770,33 @@ def from_mar_archive(cls, mar_archive: str) -> LanguageGraph:
...
+            return {}
+
+        ttc = self.predef_ttcs.get(str(ttc_entry.get('name')))
```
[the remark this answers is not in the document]
andrewbwm: So, I've been looking into this because the suggestion itself makes a lot of sense to me, the problem is you can't really implement it in a nice way, which might mean that we leave it as is.

The TTC distribution doesn't tell you that it is a predefined one when you get the dictionary for it, it just has a name, type which is function like any other one, and no arguments. And you could be able to infer that if it has no arguments it is a predefined value, but I'm not sure if that is always the case. The probability handling would complain about it in the MAL simulator if ever try to run it, so maybe that's ok.

PR #130, opened 2025-04-29
Allow attacker id in model dict to be a string

nkakouros:
This is to keep id handling on par given that an asset id can be either an interger or a string.


nkakouros:
This is a minimal change, can be merged ASAP.

PR #131, opened 2025-05-30
Langgraph walkthrough - typehinting, docstringing and simplifications

mrkickling:
This is me walking through the Language Graph and changing things the linter tells me about and adding typehints.

I also added/altered docstrings and made some simplifications of small parts of the code.

I added a couple of TODO comments I want to fix before this merge, mostly for docstrings and 'type errors'.

Some mypy is failing now because I added typehinting and we need to abide by those, also TODO before merge.

### DIFF
--- a/maltoolbox/language/languagegraph.py
+++ b/maltoolbox/language/languagegraph.py
@@ -49,7 +49,7 @@ class Detector:
 
 
 class Context(dict):
-    """TODO: What is a context?"""
+    """Context is part of detectors to provide meta data about attackers"""
     def __init__(self, context) -> None:
         super().__init__(context)
         self._context_dict = context
@@ -222,6 +222,7 @@ class LanguageGraphAsset:
 
 @dataclass
 class LanguageGraphAssociationField:
+    """A field in an association"""
     asset: LanguageGraphAsset
     fieldname: str
     minimum: int


maltoolbox/language/languagegraph.py
```diff
@@ -1052,6 +1076,9 @@ def process_step_expression(self,
                 var_name = step_expression['name']
```
mrkickling, line 1076: @andrewbwm I need some help here

maltoolbox/language/languagegraph.py
```diff
@@ -925,7 +920,8 @@ def _from_dict(cls, serialized_graph: dict) -> LanguageGraph:
                                 expr_chain_dict,
                                 lang_graph
                             )
-                            attack_step.own_requires.append(expr_chain)
+                            if expr_chain:
+                                attack_step.own_requires.append(expr_chain)
```
andrewbwm, line 924: Shouldn't this just send out an exception when it fails to generate the step expression instead?

maltoolbox/language/languagegraph.py
```diff
@@ -1585,29 +1582,29 @@ def _generate_graph(self) -> None:
...
-                        result_expr_chain, \
-                        _ = \
+                        _, result_expr_chain, _ = \
                             self.process_step_expression(
                                 attack_step.asset,
                                 None,
                                 step_expression
                             )
+                        # TODO: result_expr_chain can be None
                         attack_step.own_requires.append(result_expr_chain)
```
andrewbwm, line 1595: Not for a requirement it cannot. The only time you can get a None expression chain is if the attack step you are trying to link is local to the same asset as the originating attack step and requirements do accept take attack step names at all.

PR #133, opened 2025-05-30
Do more expensive check only if needed

nkakouros:


mrkickling:
Moved to simulator and solved there instead

PR #139, opened 2025-06-30
Update apriori propagation methods 

andrewbwm:
Update apriori propagation methods  to adequately handle both boolean values for changes and return set of nodes whose states were altered during propagation.

This PR:
- allows propagate to be used for both `False` and `True` property values
- simplify and optimize some of the logic operations being performed
- return changed nodes set from propagate methods
- ~~there is one very minor optimization/design flaw I left in there, because it's not actually worth fixing, the performance would be worse and the code more confusing. But when it does the propagation it starts by setting out the child property to the node it is propagating from and then checks all of the parents(which includes the node it was propagated from). Which means that the propagation source parent is checked twice, this does not change the result in any way and removing the node from the list would almost certainly cost more and be more confusing than just leaving it as is. I can add a comment in there, but I also think the comment would just confuse people more. It just does one redundant cheap operation again sometimes, we can live with that.~~ See comment below and @mrkickling suggestions.
- adds both viability and necessity propagation tests. It technically updated the old viability test, but that one was a mess, so they're both basically brand new.

### DIFF
--- a/maltoolbox/attackgraph/analyzers/apriori.py
+++ b/maltoolbox/attackgraph/analyzers/apriori.py
@@ -42,16 +42,12 @@ def propagate_viability_from_node(node: AttackGraphNode) -> set[AttackGraphNode]
     for child in node.children:
         original_value = child.is_viable
         if child.type == 'or':
-            child.is_viable = node.is_viable
-            parents = child.parents.copy()
-            while not child.is_viable and parents:
-                child.is_viable = parents.pop().is_viable
+            child.is_viable = any(
+                parent.is_viable for parent in child.parents)
 
         elif child.type == 'and':
-            child.is_viable = node.is_viable
-            parents = child.parents.copy()
-            while child.is_viable and parents:
-                child.is_viable = parents.pop().is_viable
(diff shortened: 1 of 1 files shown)


andrewbwm:
There is a slightly less efficient, but more legible way of implementing the logical operations to determine the property values, let me know if you'd like us to go with that instead.

maltoolbox/attackgraph/analyzers/apriori.py
```diff
@@ -21,62 +21,82 @@
...
+    changed_nodes = set()
     for child in node.children:
         original_value = child.is_viable
         if child.type == 'or':
-            child.is_viable = False
-            for parent in child.parents:
-                child.is_viable = child.is_viable or parent.is_viable
-        if child.type == 'and':
-            child.is_viable = False
+            child.is_viable = node.is_viable
```
mrkickling, line 45: Do you agree that we can replace these four lines with:
`child.is_viable = any(parent.is_viable for parent in child.parents)`
?

mrkickling (approved):

PR #145, opened 2025-08-15
Bugfix ttc calc

mrkickling:


maltoolbox/language/languagegraph.py
```diff
@@ -97,12 +97,60 @@
...
+    ttc_entry   - the TTC entry to check for predefined names
+
+    Returns:
+    If the TTC entry provided contained a predefined name the TTC
+    probability distrubtion corresponding to it. Otherwise, the TTC
+    distribution provided as a parameter as is.
+    """
+
+    # step_ttc can be either: None, predefined distribution or explicit
+    # distribution. The goal is to return explicit distribution for all steps
```
andrewbwm, line 126: I don't like the phrasing here for multiple reasons in that last sentence.

Could we change it to something like "This method will convert predefined TTC names and apply defaults to ensure uniformity(explicit distributions)."

andrewbwm (approved):

PR #242, opened 2026-06-08
Standardize mal-lang download via git url

### LINKED ISSUE #241: Standardize git url usage as language file
Currently, when you use a GitHub SSH URL as the language file, it looks for a file called `main.mal` in the `/src/main/mal` subfolder; however, this does not have to be the case for all repos. We will follow the following strategy to solve this issue:
1. Go through all directories in the repo and gather all .mal files
2. If there is one sole .mal file, use it as the mal-lang
3. If there are multiple, use the one called `main.mal`
4. Else, execution fails

auroranavas:
mal-toolbox currently looks for the file `main.mal` in the /src/main/mal subfolder. This PR fixes that and follows the approach proposed in #241.

### DIFF
--- a/tests/language/test_languagegraph.py
+++ b/tests/language/test_languagegraph.py
@@ -305,14 +305,14 @@ def test_attack_graph_node_causal_mode(tmpdir):
 
 @pytest.mark.integration
 def test_load_from_git():
-    """Test that we can pickle and unpickle a language graph attack step"""
+    """Test that we can load the mal-lang through a git repo url"""
     git_url = 'git@github.com:mal-lang/coreLang.git'
     lang_graph = load_language_graph_from_file(git_url)
     assert lang_graph.assets, "Expected assets in the loaded language graph"
 
 @pytest.mark.integration
 def test_load_from_git_not_main_mal():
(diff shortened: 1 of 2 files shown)


tests/language/test_languagegraph.py
```diff
@@ -308,4 +308,18 @@ def test_load_from_git():
     """Test that we can pickle and unpickle a language graph attack step"""
     git_url = 'git@github.com:mal-lang/coreLang.git'
     lang_graph = load_language_graph_from_file(git_url)
-    assert lang_graph.assets, "Expected assets in the loaded language graph"
\ No newline at end of file
+    assert lang_graph.assets, "Expected assets in the loaded language graph"
+
+@pytest.mark.integration
+def test_load_from_git_not_main_mal():
+    """Test that we can pickle and unpickle a language graph attack step"""
```
mrkickling: I see this is an old docstring, also on my previous test. Can you update it on both tests to say something like "Test that we can load languaged from git url"

mrkickling:
Looks great! :tada:

mrkickling (approved):

PR #244, opened 2026-07-22
DynaMAL Implementation

sandorstormen:
This PR creates an MVP for implementing [DynaMAL](https://doi.org/10.1145/3664476.3664508) in the toolbox. DynaMAL functionality is implemented as an addition to the existing MAL functionality.

The main change to the `LanguageGraph` is the addition of `LanguageGraphModelEffect` representing modifications to the instance model that are made during runtime. Each `LanguageGraphAttackStep` has the two fields `own_additive_model_effects` and `own_subtractive_model_effects` with a list of `LanguageGraphModelEffect`. Since these are dynamic in nature, the model modifications can not be calculated before runtime. Therefore `LanguageGraphModelEffect` includes information on which associations to traverse to be able to generate the correct additive/subtractive effects.

## TODO
- [ ] Figure out which asset the target asset expressions in a dynamic sentence should start from.
  - Implemented now as 1) the attack steps asset and 2) the base asset.


sandorstormen:
**Ambiguity in Assoc Traversal for Targets in a Dynamic Sentence**

Reading wiperLang, I think I have noticed some ambiguities in how the DynaMAL grammar is used. For `Wiper:exfiltrate`, the target expression `~victim.inet.hosts[C2Server].data` refers to an association which `Wiper`/`Malware` has. But the base (`victim.data`) produces a set of `Data` assets. So I assume there is an implied `self` at the start of `~victim.inet.hosts[C2Server].data`.

I think the ambiguity comes from how @viktorengstrom interprets target expressions in a dynamic sentence. First off, `self` is always implied at the start of a base/target. But, targets may also reference associations based on the set of assets that the base evaluates to. So targets may both reference `self.x` and `base.x`. I suggest we do away with this ambiguity and always write `self.x` in targets for a dynamic sentence. In all other cases, we assume `base.x` for targets in a dynamic sentence.

Just to be clear with my terminolog, here is some pseudo-regex of what I call the different parts of a dynamic sentence : `[A|R]> base / target (^ target)?`

@viktorengstrom Have I misunderstood anything? Does the field references in a target expression start with the base set of assets or does it start with `self`?

viktorengstrom:
> @viktorengstrom Have I misunderstood anything? Does the field references in a target expression start with the base set of assets or does it start with self?

**TL;DR:**

DynaMAL assumes that its sentences begin with an implied `self.` fragment by default.

The `self` keyword is a reserved role identifier that points back to the currently held set of assets within a sentence. When used at the very start of a sentence before anything else, _self_ resolves to the instigating asset or asset class.

In language graphs, targets start from the asset class(es) inferred by the base by default. The exception is link/edge addition where the target reverts to the instigating asset class. This is because of how link addition is defined as opposed to the other dynamic operations, i.e., pairing two existing objects instead of adding/removing starting from existing objects.

How the base is handled at the separator (`/`) is referred to as _carrying_ or the _carrying rules_.

We can separate between the carried/initial base and the current set of assets within a given sentence.

The _instigating asset class_ is the class that contains the definition of the attack step.

The _instigating asset_ is the instantiated object from which an attack step is executed later on.

I think this is basically what I cover in the essay below.

---

If I understand correctly, the post above touches on a few things:

- The assumed self-collect at the start of sentences.
- The carrying rules between the base and the target at the separator.
- The definition of edge/link addition as opposed to the other dynamic operations.

>The assumed self-collect at the start of sentences.

Starting with the easy one, every DynaMAL sentence has an implied `self.` fragment at the start by default, including the base and target sentences.

At the start of the base or a `->`-sentence, `self` resolves to the instigating asset or asset class, depending on the context.

At the start of a target, `self` resolves to whatever was carried from the base by default. However, link addition is an exception, which I'll unpack below.

To have it on record in this thread too, `self` is a reserved role identifier that points back to the currently held set of assets. Since expressions are assumed to start from the instigating asset by default, the initial and implied `self` typically resolves to the instigating asset.

> The carrying rules between the base and the target at the separator.

In this context, carrying refers to how the base is handled when we switch from the base to the target sentence at the `/`-separator.

The default rule is to carry the base, i.e., whatever assets the base collected, we forward them to the target. At this point, the implied `self` at the target will resolve to the base, essentially.

The one exception to the carrying rule above occurs during link addition, where the target resets back to the original instigating asset.

On the implementation side, we would talk about:

- The instigating asset.
- The carried/initial base.
- The current set of assets [in given a sentence].

> The definition of edge/link addition as opposed to the other dynamic operations.

All of the above basically comes down to the definition of link addition as opposed to the other dynamic operations.

Starting off, the other dynamic operations basically declare how to add or remove something starting from an existing object, specifically the base. 

Link addition instead pairs two existing objects, specifically a base object and a target object. Note that pairing does not assume pairing from a given object like the other operations above.

To reach and pair existing objects, the base is _not_ carried during link addition by default. For one, this is to ensure that the target does not have to reverse the base to get back to the instigating asset. It is also not guaranteed that the object you want to pair is easily reachable from the base.

Similarly, the link addition context may depend on the instigating asset rather than the remote base asset, so carrying the base would cause a context switch that could be difficult to deal with. For example, the cloud modeling stuff I did typically relied on analyzing the permissions applied to the instigating asset. If the base carried, then that context would be lost, meaning it would be difficult to reverse and infer what objects I'm allowed to connect.

That said, I'll be the first to admit that the syntax for link addition has always been counter-intuitive and subject to some monkey business.

sandorstormen:
Thank you for such a detailed response @viktorengstrom! I will make sure to treat `self` (implied or explicit) in the appropriate manner then.

mathiasekstedt (approved):
LGTM

